PR #13, opened 2020-02-20
switch licenses to mit-apl

koivunej:
see related https://github.com/ipfs-rust/rust-ipfs/pull/50.

There is only a single contributor before this so probably ok just to merge.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,7 @@ name = "libipld"
 version = "0.1.0"
 authors = ["David Craven <david@craven.ch>"]
 edition = "2018"
+license = "MIT OR Apache-2.0"
 
 [dependencies]
 async-std = "0.99"
--- a/LICENSE-APACHE
+++ b/LICENSE-APACHE
@@ -1,0 +1,201 @@
+                              Apache License
+                        Version 2.0, January 2004
+                     https://www.apache.org/licenses/LICENSE-2.0
+
+TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+1. Definitions.
+
+   "License" shall mean the terms and conditions for use, reproduction,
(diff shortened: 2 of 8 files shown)


vmx:
You were faster then I was, I also had the PR already prepared but haven't pushed it yet.

For those wondering why, I'd like to quote from https://github.com/ipfs/js-ipfs/issues/2624, for the reasons for dual-licensing under MIT/Apache2: 

> This has two major benefits:
>  - There are concerns in the open source community about whether the MIT license leaves users vulnerable to patent infringement claims. We think the pure legal risk is small, but the way the open source community interacts with our project is really important. It makes sense to pick the license that makes the largest number of people comfortable.
> - There's now no reason to adopt a separate DCO, since the Apache-2 license grant addresses the same issue. 
>
> Why use a dual license, instead of just Apache-2? The Apache-2 license is incompatible with the GPLv2 license, which includes things like the Linux kernel. With a dual license, GPLv2 projects can just use the MIT license instead. Our goal is to make our software available to as many projects as possible, so we'd rather adopt a licensing scheme that doesn't exclude anyone.

vmx:
Best would be if @dvc94ch could approve this PR and perhaps even comment:

> I license past and future contributions under the dual MIT/Apache-2.0 license, allowing licensees to choose either at their option.

dvc94ch:
We discussed on riot, I'm not sure how dual licensing provides patent protection, since I can use rust-ipfs under the MIT license and still sue rust-ipfs for patent infringement. But as I said, it's ok with me

vmx (changes requested):
@koivunej Could you please also add the `license = "MIT OR Apache-2.0"` to all crates in this repo?

vmx (approved):

koivunej:
All Cargo.tomls now have the license, failed with my sed magic the first time but now they seem to be ok :)

Issue #8, opened 2020-01-29
Making codec calls sync


vmx:
Based on https://github.com/ipld/ipld/issues/83#issuecomment-579429957 it seems that the codec encoder/decoder should be made sync again.

This would mean that the Codec Trait would change its [encode and decode methds](https://github.com/ipfs-rust/rust-ipld/blob/master/base/src/codec.rs#L18-L20) from `async` to sync
(the CBOR implementation of it is here: https://github.com/ipfs-rust/rust-ipld/blob/195df89087a31397c4a37a07e0694ded8ce0ad02/dag-cbor/src/lib.rs#L26-L34)

@dvc94ch is that correct?

Gozala:
I can take this

dvc94ch:
Before making any more changes we should probably start by adding some benchmarks, and then validate that this does make a huge difference.

vmx:
I'm always in favour of having benchmarks, though in this case I don't think it has a high priority. I would accept a PR making things sync even without benchmarks. Performance surely is a concern at one point, but I don't think we are there yet.

dvc94ch:
Well it is very useful to have an idea what takes a lot of time and what doesn't. There are huge surprises and regularly checking is advisable. For example using base58 for cids for anything other than displaying to the user is extremely unadvisable. Something I'd not have thought before looking at the flamegraph.

Gozala:
Maybe it is a good time to ask what is the motivation for switching to sync. I was assuming that async lead to poor performance. If that is not the case it would be to know the motivation. Maybe experience of working with async is poorer ? 

Anyway it seems like it’s good to know what the goal is so we can verify afterwards that it is accomplished. If it is performance we should definitely check that numbers are improved, if it is something else let’s make sure that our assumption about this change are grounded in some evidence

vmx:
To me the motivation for having it sync would be to have easier to deal with code, if the performance is not worse by an order of a magnitude compared to the async one.

dvc94ch:
I expect there to be an order of magnitude improvement, I think the problem is that the compiler doesn't support async traits yet, it missed a lot of inlining/optimization improvements. I'll have to look into it again, will probably get around to it soon.

dvc94ch:
It being easier to deal with code is not a reason. Using async means that the parser works on a stream directly. Sending ipld over tcp/uds would require delimiting the data, buffering it and then parsing the buffer. That means that it's slightly more work for the programmer.

Gozala:
[...]

PR #19, opened 2020-02-23
Switch to sync API

Gozala:
As per #8 updating implementation to use sync API.


base/src/error.rs
```diff
@@ -67,6 +68,13 @@ pub enum BlockError {
     /// Io error.
     #[fail(display = "{}", _0)]
     Io(std::io::Error),
+    // RwLock error
+    #[fail(display = "{}", _0)]
+    RwLockReadGuard(failure::Error),
+    #[fail(display = "{}", _0)]
+    PoisonError(failure::Error),
+    #[fail(display = "{}", _0)]
+    RwLockWriteGuard(failure::Error),
```
Gozala, line 77: I do not understand how previous implementation avoided this since `BufStore` uses `RWLock`, I suppose `blocks.read().await` somehow avoids having to deal with `RwLockReadGuard` but I don't understand how.
I would love if someone could explain that to me.

I am also not sure if there is a better way to go about it.

base/src/error.rs
```diff
@@ -86,3 +94,21 @@ impl From<cid::Error> for BlockError {
         Self::Cid(err)
     }
 }
+
+impl<T: std::fmt::Debug> From<RwLockReadGuard<'_, T>> for BlockError {
```
Gozala, line 98: I did not wanted to complicate `BlockError` by introducing parameters. I could not figure out a better way to do this, please help 😭

dag-cbor-derive/Cargo.toml
```diff
@@ -8,14 +8,13 @@ edition = "2018"
...
 [dependencies]
-async-trait = "0.1"
 proc-macro2 = "1.0"
 quote = "1.0"
 syn = "1.0"
 synstructure = "0.12.0"
 
 [dev-dependencies]
-async-std = "0.99"
+libipld = { path = "../" }
```
Gozala, line 17: From what I can tell `dag_cbor_derive` uses `libipld` which it can't find without this.

dag-cbor/src/encode.rs
```diff
@@ -1,292 +1,269 @@
...
-pub use futures_io::AsyncWrite as Write;
 use libipld_base::cid::Cid;
 use libipld_base::ipld::Ipld;
 use std::collections::BTreeMap;
+pub use std::io::Write;
 
 #[inline]
-pub async fn write_null<W: Write + Unpin + Send>(w: &mut W) -> Result<()> {
-    w.write_all(&[0xf6]).await?;
+pub fn write_null<W: Write + Unpin + Send>(w: &mut W) -> Result<()> {
```
dvc94ch, line 10: The Unpin + Send traitbounds are not required anymore

dag-cbor/src/encode.rs
```diff
@@ -1,292 +1,269 @@
...
         for value in self {
-            value.write_cbor(w).await?;
+            value.write_cbor(w)?;
         }
         Ok(())
     }
 }
 
-#[async_trait]
 impl<T: WriteCbor + Send + Sync + 'static> WriteCbor for BTreeMap<String, T> {
```
dvc94ch, line 243: Send + Sync traitbounds are not required anymore

src/store.rs
```diff
@@ -5,83 +5,80 @@ use crate::error::Result;
...
-use futures::join;
 use std::collections::HashMap;
 use std::mem;
 use std::path::Path;
 use std::sync::Arc;
+use std::sync::RwLock;
 
 /// Implementable by ipld storage backends.
-#[async_trait]
+
```
dvc94ch, line 17: The Store should remain async.

dvc94ch: Or at least it's a completely different issue and unrelated to #19. ~It's also not on the hotpath and unlikely to have an large inpact on performance imo~

base/src/error.rs
dvc94ch: the store was always async, it's merely the codec I meant in #19

dvc94ch:
Did you notice any difference in the benchmarks, as it seems some already exist?

dag-cbor-derive/Cargo.toml
dvc94ch: the examples are mainly used here:
```rust
#[cfg(test)]
mod tests {
    #[test]
    fn test() {
        let t = trybuild::TestCases::new();
        t.pass("examples/basic.rs");
        t.pass("examples/name_attr.rs");
        //t.pass("examples/repr_attr.rs");
    }
}
```

base/src/error.rs
Gozala: Got it. @dvc94ch should I add a change to make store async again ? (Asking because I noticed your other pull requests are based off this).

base/src/error.rs
Gozala: @dvc94ch This is no longer will be a problem as store will go back to being async, but would you mind still commenting what would have being a right way to go about it ?

dag-cbor-derive/Cargo.toml
Gozala: I do get it but type-checker seemed to complain about `libipld` not being in the scope. From what I gathered it was due to the fact that `libipld` was not resolved in the code below:

```rust
fn dag_cbor_derive(s: Structure) -> TokenStream {
    let write_cbor = gen::write_cbor(&s);
    let read_cbor = gen::read_cbor(&s);
    s.gen_impl(quote! {
        use libipld::cbor::IpldError;
        use libipld::cbor::{ReadCbor, WriteCbor, CborResult as Result};
        use libipld::cbor::encode::{write_u64, Write};
        use libipld::cbor::decode::{read_u8, read_key, Read};

        gen impl WriteCbor for @Self {
            #write_cbor
        }

        gen impl ReadCbor for @Self {
            #read_cbor
        }
    })
}
```

Gozala:
@dvc94ch it appears that recent changes in https://github.com/multiformats/rust-multibase are incompatible with https://github.com/dvc94ch/rust-cid/tree/cli as I'm getting following:

```
/Users/gozala/.cargo/git/checkouts/rust-cid-857ab8699b5b856b/52c079b/src/lib.rs:77:50
   |
77 |         let mut string = multibase::encode(Base::Base58btc, &self.hash.as_ref());
   |                                                  ^^^^^^^^^
   |                                                  |
   |                                                  variant or associated item not found in `multibase::base::Base`
   |                                                  help: there is a variant with a similar name: `Base58Btc`

error[E0599]: no variant or associated item named `Base58btc` found for enum `multibase::base::Base` in the current 
```

base/src/error.rs
dvc94ch: I removed the part that makes the store sync from the PR. I think my PR superceedes this one and we can close it, if you're ok with it.

Gozala: Ah ok, I must have missed those changes. 👍

Gozala:
Closing in favor of #20

dag-cbor-derive/Cargo.toml
dvc94ch: No it's a mistake your completely right, I just figured that that's the reason I didn't notice it.

base/src/error.rs
dvc94ch: The correct way to go about it would have been:
 ```rust
impl From<std::sync::PoisonError> for BlockError {
    fn from(error: std::sync::PoisonError) -> Self {
        Self::PoisonError(error)
    }
}
```
and the three cases would have collapsed into one if I'm not mistaken.

PR #21, opened 2020-02-23
Ci

dvc94ch:
This is wip. The multihash fork needs to be resolved as fasthash doesn't build on windows.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,107 @@
+on: [push, pull_request]
+
+name: Rust IPLD
+
+jobs:
+  ci:
+    strategy:
+      fail-fast: false
+      matrix:
+        toolchain:
+        - rust: stable
+        #- rust: nightly
+        platform:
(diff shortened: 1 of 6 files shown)


koivunej:
Since we have to do similar for other repos, an I am quite new to github workflows, could you link the source you used for these workflows unless you wrote them yourself? I had been looking at https://github.com/actions-rs/example/blob/master/.github/workflows/quickstart.yml as a starting point, but in the ci.yml in this draft at least the actions come from different repository. Do you have any handle on which one of these is more recent/suggested/most official? :)

dvc94ch:
I used `actions-rs` first, but it seems more complicated and I think I had some issues with caching the cargo folder, but I don't remember what it was.

vmx:
As an inspiration: some multiformats repos were moved from Travis to GitHub Actions, e.g. https://github.com/multiformats/rust-multihash/pull/57 and seems to work well: https://github.com/multiformats/rust-multihash/actions/runs/48816477

dvc94ch:
Superceeded by #22

PR #22, opened 2020-03-06
Use released deps.

dvc94ch:
Fixes a few issues and precursor to adding it as a dependency to rust-ipfs


vmx (commented):
Looks good, I'd wait for the new rust-cid release as it will also contain the getters of the CID.

dvc94ch:
can we get the arbitraries in too? Doesn't really matter, I'll just impl it myself

vmx (commented):
The change looks good. I'd just wait for a new release of rust-cid as it would break things again (due to using getters for the CID.

vmx:
I released rust-cid 0.4.

PR #23, opened 2020-03-06
chore: add CI

vmx:
This CI file was copied from the https://github.com/multiformats/rust-multihash.

I added the `--workspace` flag to the tests, so that the tests in all workspaces
are run.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,67 @@
+on: [push, pull_request]
+
+name: build
+
+jobs:
+  build:
+    name: Lints and Test
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


vmx:
This file also adds support for code coverage reports. This would mean that we need to enable codecov.io for this org. I'm not part of the "we must have 100% test coverage camp", but I think having some indication is useful.

dvc94ch:
What's wrong with the other ci build? It already includes everything required for ci testing on android/ios/wasm. Just needs to be enabled.

vmx:
Oh I didn't see that you added CI to https://github.com/ipfs-rust/rust-ipld/pull/22. I just saw that there was no CI so I added it. I'm happy to use the other file (though I would've preferred if it would've been in a separate PR).

PR #25, opened 2020-03-09
Ipfs integration

dvc94ch:
Adds dag protobuf support for ipfs


base/src/error.rs
```diff
@@ -50,31 +50,31 @@ impl From<core::convert::Infallible> for IpldError {
 }
 
 /// Block error.
-#[derive(Debug, Fail)]
+#[derive(Debug, Error)]
 pub enum BlockError {
     /// Block exceeds MAX_BLOCK_SIZE.
-    #[fail(display = "Block size {} exceeds MAX_BLOCK_SIZE.", _0)]
+    #[error("Block size {0} exceeds MAX_BLOCK_SIZE.")]
     BlockToLarge(usize),
```
aphelionz: BlockTooLarge*

Cargo.toml
```diff
@@ -12,12 +12,13 @@ byteorder = "1.3.4"
...
 dag-cbor = { path = "dag-cbor" }
 dag-cbor-derive = { path = "dag-cbor-derive" }
-failure = "0.1.7"
+dag-pb = { path = "dag-pb", optional = true }
 futures = "0.3.4"
 libipld-base = { path = "base" }
 libipld-macro = { path = "macro" }
 multibase = "0.8.0"
 multihash = "0.10.1"
+thiserror = "1.0.11"
```
koivunej, line 21: I wonder if this could be "1.0" or is there something behind pinning a "past" 1.0 version?

dag-cbor/Cargo.toml
```diff
@@ -7,8 +7,8 @@ license = "MIT OR Apache-2.0"
 
 [dependencies]
 byteorder = "1.3.4"
-failure = "0.1.7"
 libipld-base = { path = "../base" }
+thiserror = "1.0.11"
```
koivunej, line 11: Same pinning here.

dag-pb/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "dag-pb"
+version = "0.1.0"
+authors = ["David Craven <david@craven.ch>"]
+edition = "2018"
+
+[dependencies]
+libipld-base = { path = "../base" }
+prost = "0.6.1"
+thiserror = "1.0.11"
```
koivunej, line 10: And here.

dag-pb/src/lib.rs
```diff
@@ -0,0 +1,74 @@
...
+use core::convert::TryInto;
+use libipld_base::cid;
+use libipld_base::codec::Codec;
+use libipld_base::error::{BlockError, IpldError};
+use libipld_base::ipld::Ipld;
+use thiserror::Error;
+
+mod codec;
+
+/// Protobuf codec.
```
koivunej, line 12: Could link to spec here https://github.com/ipld/specs/blob/master/block-layer/codecs/dag-pb.md

dag-pb/src/dag_pb.proto
```diff
@@ -0,0 +1,21 @@
...
+  uint64 Tsize = 3;
+}
+
+// An IPFS MerkleDAG Node
+message PBNode {
+  // refs to other objects
+  repeated PBLink Links = 2;
+  // opaque user data
+  bytes Data = 1;
+}
```
koivunej, line 22: This file itself has been updated in the spec with different comments: https://github.com/ipld/specs/blob/master/block-layer/codecs/dag-pb.md

koivunej (commented):
This is looking good.

Cargo.toml
dvc94ch: patch versions should be compatible. If the crates semver correctly, the best practice imo is to use the version that it was compiled against, like `cargo update` or `cargo upgrade` do.

koivunej: Oki, I'll resolve these then.

vmx (commented):
Looks good! Running `cargo fmt` and it should be ready to go.

vmx (approved):
Once `cargo fmt` is applied it's ready to be merged.

PR #26, opened 2020-03-11
fix: ci stop using actions/checkout@v1

koivunej:
with the https://github.com/ipfs-rust/rust-ipfs/pull/85 there were first some spurious failures on windows (probably transient gh infra issues) and later we noticed that the checkout action on lint-rust and readme-doctest was failing.

I noticed the checkout action on the failing jobs were using an older version of the failing action than the matrix action which succeeded. tried to upgrade that and it worked, didn't look further.

Cherry-picked from 8795a15 on eqlabs/rust-ipfs which is now https://github.com/ipfs-rust/rust-ipfs/pull/87

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,7 +95,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Checkout sources
-      uses: actions/checkout@v1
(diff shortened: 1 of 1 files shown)


vmx (approved):

koivunej:
Going ahead, same as the other PR in rust-ipfs.

PR #28, opened 2020-04-04
Dag json

### LINKED ISSUE #17: Add dag-json implementation
Once #8 is complete, I would like to revive https://github.com/vmx/rust-ipld/pull/2 to add dag-json support here.

We did have a lengthy conversations about where we should or should not use serde. My personal opinion had being let's use it where it's usable given it's well tested and time proved, we can always rewrite internals (to rip it off) after the fact.

If it does not feel reasonable please let me know.

dvc94ch:
Closes #17, based on @Gozala 's implementation

### DIFF
--- a/base/src/lib.rs
+++ b/base/src/lib.rs
@@ -5,3 +5,5 @@ pub mod hash;
 pub mod ipld;
 
 pub use cid;
+pub use multibase;
+pub use multihash;
--- a/dag-json/src/codec.rs
+++ b/dag-json/src/codec.rs
@@ -1,0 +1,269 @@
+use crate::JsonError;
+use core::convert::TryFrom;
+use libipld_base::cid::Cid;
+use libipld_base::ipld::Ipld;
+use serde::{de, ser, Deserialize, Serialize};
+use serde_json::ser::Serializer;
+use std::collections::BTreeMap;
+use std::fmt;
+use std::iter::FromIterator;
+
+const LINK_KEY: &str = "/";
+
+pub fn encode(ipld: &Ipld) -> Result<Box<[u8]>, JsonError> {
+    let mut writer = Vec::with_capacity(128);
+    let mut ser = Serializer::new(&mut writer);
+    serialize(&ipld, &mut ser)?;
+    Ok(writer.into_boxed_slice())
+}
+
+pub fn decode(data: &[u8]) -> Result<Ipld, JsonError> {
+    let mut de = serde_json::Deserializer::from_slice(&data);
+    Ok(deserialize(&mut de)?)
+}
+
+fn serialize<S>(ipld: &Ipld, ser: S) -> Result<S::Ok, S::Error>
+where
+    S: ser::Serializer,
+{
+    match &ipld {
+        Ipld::Null => ser.serialize_none(),
+        Ipld::Bool(bool) => ser.serialize_bool(*bool),
+        Ipld::Integer(i128) => ser.serialize_i128(*i128),
+        Ipld::Float(f64) => ser.serialize_f64(*f64),
+        Ipld::String(string) => ser.serialize_str(&string),
+        Ipld::Bytes(bytes) => ser.serialize_bytes(&bytes),
+        Ipld::List(list) => {
+            let wrapped = list.iter().map(|ipld| Wrapper(ipld));
+            ser.collect_seq(wrapped)
+        }
+        Ipld::Map(map) => {
+            let wrapped = map.iter().map(|(key, ipld)| (key, Wrapper(ipld)));
+            ser.collect_map(wrapped)
+        }
+        Ipld::Link(link) => {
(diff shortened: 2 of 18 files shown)


dag-json/src/codec.rs
```diff
@@ -0,0 +1,269 @@
...
+        // JSON Object represents IPLD Link if it is `{ "/": "...." }` therefor
+        // we valiadet if that is the case here.
+        if let Some((key, WrapperOwned(Ipld::String(value)))) = values.first() {
+            if key == LINK_KEY && values.len() == 1 {
+                // TODO: Find out what is the expected behavior in cases where
+                // value is not a valid CID (or base64 endode string here). For
+                // now treat it as some other JSON Object.
+                let link = base64::decode(value).unwrap();
+                let cid = Cid::try_from(link).unwrap();
+                return Ok(Ipld::Link(cid));
```
dvc94ch, line 194: shouldn't panic, but probably requires quite a bit more code. anyone familiar with serde? @Gozala

koivunej:
The failure with aarch64-linux-android is a transient failure to `apt-get update`.

dag-json/src/codec.rs
koivunej: If the `&'de str` could be acquired for this value then it would be a matter of returning https://docs.rs/serde/1.0.106/serde/de/trait.Error.html#method.invalid_value or `invalid_length` for these.

In the mean time the following:

```rust
                let link = base64::decode(value)
                    .map_err(|e| serde::de::Error::custom(e))
                    .and_then(|link| Cid::try_from(link).map_err(|e| serde::de::Error::custom(e)))?;

                return Ok(Ipld::Link(link));
```

passes the test:

```rust
    #[test]
    fn decode_invalid() {
        let input = r#"{ "/": "invalidcid" }"#;

        decode(input.as_bytes()).unwrap_err();
    }
```

The error message just is quite confusing: `Json(Error("Invalid last symbol 100, offset 9.", line: 1, column: 21))`

koivunej: Ah of course, when the `&'de str` is not available we can just have `'static` errors, this gives a nicer: `invalid base64 encoding in link "invalidcid": Invalid last symbol 100, offset 9. at line 1 column 21`:

```
                #[derive(Debug)]
                enum InvalidLink {
                    InvalidEncoding(String, base64::DecodeError),
                    InvalidCid(String, libipld_base::cid::Error),
                }

                impl std::fmt::Display for InvalidLink {
                    fn fmt(&self, fmt: &mut std::fmt::Formatter) -> std::fmt::Result {
                        match self {
                            InvalidLink::InvalidEncoding(s, e) => {
                                write!(fmt, "invalid base64 encoding in link {:?}: {}", s, e)
                            },
                            InvalidLink::InvalidCid(s, e) => {
                                write!(fmt, "invalid cid in link {:?}: {}", s, e)
                            }
                        }
                    }
                }

                impl std::error::Error for InvalidLink {}

                let raw_cid = match base64::decode(value) {
                    Ok(bytes) => bytes,
                    Err(e) => {
                        if let (_, WrapperOwned(Ipld::String(value))) = values.swap_remove(0) {
                            return Err(serde::de::Error::custom(InvalidLink::InvalidEncoding(value, e)));
                        } else {
                            unreachable!("IPLD variant already checked");
                        }
                    }
                };

                let cid = match Cid::try_from(raw_cid) {
                    Ok(cid) => cid,
                    Err(e) => {
                        if let (_, WrapperOwned(Ipld::String(value))) = values.swap_remove(0) {
                            return Err(serde::de::Error::custom(InvalidLink::InvalidCid(value, e)));
                        } else {
                            unreachable!("IPLD variant already checked");
                        }
                    }
                };

                return Ok(Ipld::Link(cid));
```

koivunej: PR'd: https://github.com/dvc94ch/rust-ipld/pull/1

base/src/lib.rs
```diff
@@ -5,3 +5,5 @@ pub mod hash;
 pub mod ipld;
 
 pub use cid;
+pub use multibase;
+pub use multihash;
```
koivunej, line 9: I think this is a good idea, it'll help with using the same versions downstream.

vmx (approved):
Thanks! In the long run I'd like to move away from Serde, but that's a totally separate issue and until then, it's surely great to have dag-json support.

It would be great to have some tests. Sadly neither the [Go implementation](https://github.com/ipld/go-ipld-prime/blob/44af114587ec49c52336b8bb577b8f6062d8646d/codec/dagjson/roundtrip_test.go), not the [JS implementation](https://github.com/ipld/js-dag-json/blob/3862400e5f82b91a9caffaa645ebd287975703ef/test/dag-json.spec.js) has much tests.

I'd suggest creating file based fixtures that all implementations could then use. But that surely is for a separate PR.

PR #30, opened 2020-04-07
Fix cbor cid serde

koivunej:
The zero byte prefix inside the length delimited byte array was not omitted or added.

### DIFF
--- a/dag-cbor/src/decode.rs
+++ b/dag-cbor/src/decode.rs
@@ -106,7 +106,8 @@ pub fn read_link<R: Read>(r: &mut R) -> Result<Cid> {
     }
     let len = read_u8(r)?;
     let bytes = read_bytes(r, len as usize)?;
-    Ok(Cid::try_from(bytes)?)
+    // skip the first byte per https://github.com/ipld/specs/blob/master/block-layer/codecs/dag-cbor.md#links
+    Ok(Cid::try_from(&bytes[1..])?)
 }
 
 pub trait ReadCbor: Sized {
--- a/dag-cbor/src/decode.rs
+++ b/dag-cbor/src/decode.rs
@@ -105,6 +105,9 @@ pub fn read_link<R: Read>(r: &mut R) -> Result<Cid> {
         return Err(CborError::UnknownTag);
     }
     let len = read_u8(r)?;
+    if len == 0 {
+        return Err(CborError::LengthOutOfRange);
+    }
     let bytes = read_bytes(r, len as usize)?;
     if bytes[0] != 0 {
         return Err(CborError::InvalidCidPrefix(bytes[0]));
--- a/dag-cbor/src/decode.rs
+++ b/dag-cbor/src/decode.rs
@@ -106,7 +106,12 @@ pub fn read_link<R: Read>(r: &mut R) -> Result<Cid> {
     }
     let len = read_u8(r)?;
     let bytes = read_bytes(r, len as usize)?;
-    // skip the first byte per https://github.com/ipld/specs/blob/master/block-layer/codecs/dag-cbor.md#links
+    if bytes[0] != 0 {
+        return Err(CborError::InvalidCidPrefix(bytes[0]));
+    }
+
+    // skip the first byte per
+    // https://github.com/ipld/specs/blob/master/block-layer/codecs/dag-cbor.md#links
     Ok(Cid::try_from(&bytes[1..])?)
 }
 
--- a/dag-cbor/src/encode.rs
+++ b/dag-cbor/src/encode.rs
@@ -218,8 +218,11 @@ impl WriteCbor for Cid {
     #[inline]
     fn write_cbor<W: Write>(&self, w: &mut W) -> Result<()> {
         write_tag(w, 42)?;
+        // insert zero byte per https://github.com/ipld/specs/blob/master/block-layer/codecs/dag-cbor.md#links
         let bytes = self.to_bytes();
-        bytes.as_slice().write_cbor(w)?;
+        write_u64(w, 2, (bytes.len() + 1) as u64)?;
+        w.write_all(&[0])?;
+        w.write_all(&bytes)?;
         Ok(())
     }
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -398,6 +398,7 @@ name = "dag-cbor"
 version = "0.1.0"
 dependencies = [
  "byteorder",
+ "hex",
  "libipld-base",
  "libipld-macro",
  "multihash",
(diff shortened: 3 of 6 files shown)


dag-cbor/src/decode.rs
```diff
@@ -106,7 +106,8 @@ pub fn read_link<R: Read>(r: &mut R) -> Result<Cid> {
     }
     let len = read_u8(r)?;
     let bytes = read_bytes(r, len as usize)?;
-    Ok(Cid::try_from(bytes)?)
+    // skip the first byte per https://github.com/ipld/specs/blob/master/block-layer/codecs/dag-cbor.md#links
+    Ok(Cid::try_from(&bytes[1..])?)
```
koivunej, line 118: the first byte should probably be checked to be zero, I'll add that

dag-cbor/src/decode.rs
```diff
@@ -106,7 +106,13 @@ pub fn read_link<R: Read>(r: &mut R) -> Result<Cid> {
     }
     let len = read_u8(r)?;
     let bytes = read_bytes(r, len as usize)?;
-    Ok(Cid::try_from(bytes)?)
+    if bytes[0] != 0 {
```
koivunej, line 112: this can panic, fixing.

aphelionz:
LGTM

dag-cbor/src/encode.rs
```diff
@@ -218,8 +218,11 @@ impl WriteCbor for Cid {
     #[inline]
     fn write_cbor<W: Write>(&self, w: &mut W) -> Result<()> {
         write_tag(w, 42)?;
+        // insert zero byte per https://github.com/ipld/specs/blob/master/block-layer/codecs/dag-cbor.md#links
         let bytes = self.to_bytes();
-        bytes.as_slice().write_cbor(w)?;
+        write_u64(w, 2, (bytes.len() + 1) as u64)?;
```
vmx, line 223: What about not writing the bytes manually but doing something like:

```rust
        let slice: &[u8] = &[&[0], &bytes[..]].concat();
        slice.write_cbor(w)?;
```

dvc94ch:
I haven't understood the why yet. Is this part of the cbor or the dag cbor spec? Or is this a bug in the go implementation?

vmx:
@dvc94ch Part of the DagCBOR spec as linked in the source: https://github.com/ipld/specs/blob/51eb2f682f554b726021f34398843f8edf97bbd0/block-layer/codecs/dag-cbor.md#links. The CID is Multibase encoded, hence has a `0x0` prefix.

dvc94ch:
I'm pretty sure that section was added after implementing this version. We'll have to live with "historical reasons" then...

dag-cbor/src/encode.rs
koivunej: Hmm I didn't know of https://doc.rust-lang.org/std/primitive.slice.html#method.concat ... What does it do exactly? It cannot return a slice like in your example, and I would prefer a single additional call to a writer in favor of allocating a new `Vec<_>`. 

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=d6174c92b48b9fa208b3d00741aa60bf shows it returning a Vec.

vmx: I've tried it locally, all tests pass.

vmx: That's more like what happens in this case: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=5fa868c33ca5f63e310e69d60fad7527

vmx: > What does it do exactly?

I should answer your question. It concatenates slices.

aphelionz: Isn't the additional write call one fewer heap allocation?

aphelionz: Also @vmx just a side note: I'm wondering if there could be some construct for this in `rust-cid` i.e. some `Raw` format for this case?

koivunej: > It concatenates slices.

Yes, understood. But as it does it by copying them into a new `Vec`, which requires an allocation, I was thinking it shouldn't be done here. I was initially surprised by the suggestion as I didn't see the `let slice: &[u8] = &expr;`, the ampersand on the right hand side and was confused as to how could a concatenated slice be constructed so (since a slice is a pointer and length).

I guess the point could be made that `Write::write_all` can end up allocating (in case it's a `std::io::Cursor<Vec<u8>>` for example) but at least it will be amortized. When writing to a file this can cause a syscall to be made but then again, `BufWriter` is usually used to alleviate if that becomes a problem.

vmx: > Also @vmx just a side note: I'm wondering if there could be some construct for this in `rust-cid` i.e. some `Raw` format for this case?

The "raw" format is what `to_bytes()` is doing. You could use multibase to add the `0x0` prefix, but I think this overkill for this simple case.

vmx: Before getting into tech details. Would you agree that what I propose is kind of cleaner as it shows that we prefix the CID with `0x0` and we use existing methods to encode it (the one to encode slices into CBOR bytes), so it's less duplication of code?

dvc94ch: Concat will result in an additional heap allocation, I agree with @koivunej, just my 2cents...

koivunej: @vmx Yes your suggestion would definetly be less duplication.

I wager that this code doesn't have much pressure to be changed over time, unless of course the specs change, but still there are tests, so I felt confident selecting the less-allocating route.

vmx: I generally favour easier to read code and less potential bugs over micro-optimizations in case there are no benchmarks that are based on real-world performance problems. But it's fine for me to keep as it currently is. This was just a suggestion.

dvc94ch: I agree with the sentiment, however the codec is performance critical as every change to a block requires recursively reencoding with the updated hash...

dvc94ch: also I believe we do have benchmarks, so if someone would like to run them... :)

koivunej: Thanks for the suggestion. My opinion here is/was also biased because I think allocations are something which the compiler cannot optimize through (as in, it cannot ever remove an allocation) and so we should try to be mindful of those when possible. TBH I haven't really checked for a while if that is accurate.

aphelionz:
Now that that convo is resolved, ok to merge @dvc94ch ?

dvc94ch (approved):

vmx (approved):

PR #31, opened 2020-04-16
Fix clippy workspace

koivunej:
This is similar to https://github.com/ipfs-rust/rust-ipfs/pull/156, I noticed that clippy was only ran on the root crate instead of the whole workspace, tests and examples.


koivunej:
Failed check timed out, gh being busy.

Issue #27, opened 2020-04-04
Replace WriteCbor and ReadCbor.


dvc94ch:
Instead of `WriteCbor` and `ReadCbor` we should use something like `Encode<Codec::Cbor>` and `Decode<Codec::Cbor>`. On the derive side we could potentially add a generic derive that would be used for `dag-pb` and `dag-json` via an intermediate encoding to ipld and still keep the custom code for cbor that is much faster.

PR #32, opened 2020-04-27
Redesign store trait.

dvc94ch:
Based on feedback from https://github.com/ipfs-rust/rust-ipfs/issues/136 and experience with building stores, here is a proposal for a new store trait.

@vmx @koivunej @Gozala @aphelionz @rklaehn @alexjg @mikeal thoughts?


vmx:
Will IPLD heavily use this trait internally or can it be seen as another building block you might use but don't have to? E.g. IPFS will surely use this.

I'm asking as I think you should be able to use the core of IPLD without the need to have a storage.

dvc94ch:
We have a couple of traits/abstractions. Most of which need to be redesigned but not necessarily dumped.

Store abstracts over storage/network backend.
Encode/Decode abstracts over a codec. (#27)
Maybe a Block abstraction? You haven't finished your proposal yet... :)

Libraries like ipld-collections probably want to use the Store trait?

aphelionz:
Thanks for making the proposal @dvc94ch and for weighing in @vmx. I like a lot of stuff here, and have some questions.

First though, I want to make sure I have my head around the big picture i.e. where all the different abstractions live and what they should each accomplish. Here's my attempt at enumerating all the different abstractions, Maybe if we can look at the full landscape we can figure out what to do / where things should go.

1. The BlockStore trait which is implemented by FsStore and MemStore
2. The DataStores
3. The Repo< >abstraction which glues everything together between those, then
4. This one, and finally
5. Volker's pending Block proposal, though I suppose that might be slightly tangential to this.

I'm wondering what we're optimizing for (perhaps developer experience?), and towards that what combination of the following we can utilize:
1. Simple low-level traits (like perhaps a Block is just a `Box<[u8]>` to start and can be "outfitted" with traits and other fields as needed)
2. Simple higher-abstracted traits that allow users to further mix and match the features and implementations they need (i.e. not everybody would want pinning, or GC, or Aliasing)

dvc94ch:
@aphelionz the BlockStore, DataStore and Repo are currently internal details of rust-ipfs. Neither `ipfs-embed` or the `ipld-daemon` make use of those concepts.

Aliasing is essentially what ipns does, it is an important feature for most applications. But yes it's currently split in a separate trait for that reason.

If you don't want to implement gc and just remove the block on unpin, you could provide a blanket implementation. The semantics of gc should depend on if a link is strong or weak.

aphelionz:
> Neither ipfs-embed or the ipld-daemon make use of those concepts.

Sure, perhaps it could? In other words, is there anything stopping us from simply adding functionality like pinning to the Blockstore trait within Ipfs and then using it in `rust-ipld`?

In my mind the blockstore's "job" would be to handle the raw blocks while IPLD is more high-level, dealing with the schemas and linking and how everything is connected, so moving a lot of these traits and functionality into `rust-ipfs` might make more sense.

All this being said, I'm not yet a _user_ of the library so I'm wondering if there's something cumbersome about the way that would work / is done now?

dvc94ch:
The problem with that is that it would make ipld depend on rust-ipfs instead of the other way around. For my current job I'm implementing an embeddable ipfs repo that will share it's backing db with a substrate light client. I don't want to deploy two databases to a mobile phone... And certainly not a http server :)

aphelionz:
Aha! That's helpful to know. Thinking more now...

aphelionz:
So first off: totally open to the idea of creating more / multiple `ipfs-****` binaries as I believe you are with `ipfs-embed` and I think in the end `rust-ipfs` should be usable first and foremost as just a _really good Rust crate_ to use elsewhere.

As for IPLD, I'm starting to "get it," thanks for explaining more. So the design here is that `rust-ipld` should be able to use any file store that implements your `Store` trait as written. The traits live here but the implementations live elsewhere?

So as a user I'd:

```rust
struct SomeStore { /* ... */ } 
impl libipld::Store for SomeStore {
    // fn insert ,etc
}
```

If so, would that be too much work for the implementer i.e. would they have to do all the requisite encoding while implementing that trait, using `rust-ipld` functions explicitly, or does an implementation have to live in `rust-ipld` as well?

or, OR! Could this potentially be a derive to "IPLD-ify" a store? That could be interesting. Implement the base blockstore and put the IPLD derive on it to do all the encoding under the hood :laughing:

dvc94ch:
@aphelionz the initial idea was to make ipfs configurable with custom block stores and data stores, but that is independent of if ipld or anything built on top of it should be tied to a specific implementation. And I think there are a few issues that need to be solved. One of them is that rust-ipfs is currently not compatible with rust-ipld. Also the semantics of pinning and garbage collection seem somewhat broken as mentioned previously.

The proposed semantics of pinning here is to register a gc root and have the gc make sure that there are no live references. So the inteded usage would be like this:

```
insert(a)
insert(b -> a)
unpin(a)
insert(c -> a)
unpin(b)
live() == [a, c]
unpin(c)
live() == []
```

base/src/store.rs
```diff
@@ -0,0 +1,62 @@
...
+    fn gc(&self) -> StoreResult<()>;
+}
+
+/// Implemented by ipld storage backends that support multiple users.
+pub trait MultiUserStore: Store {
+    /// Pin a block.
+    ///
+    /// This creates a symlink chain from root -> path -> block. The block is unpinned by
+    /// breaking the symlink chain.
+    fn pin(&self, cid: &Cid, path: &Path) -> StoreResult<()>;
```
aphelionz: I think i see why this is here, though having `pin` and `unpin` split across two separate traits makes me a little dizzy. :sweat_smile: What other functions would be added to MultiUserStore in the future.

Would the different types of stores be more suited as an enum? Basically `rust-ipld`'s way of broadcasting outward: "these are the different types of stores I know what to do with"

dvc94ch:
implementations should not live in `rust-ipld` in general. it makes sense to have an implementation for testing purposes, but it's not intended to be used outside of testing. There are some utilities that I implemented for implementing the garbage collector or caching/buffering, but those can probably be moved somewhere else. The store trait is intended to be the "minimum" necessary api that needs to be implemented. Any ipld application that uses rust-ipld should be compatible with any store.

base/src/store.rs
```diff
@@ -0,0 +1,62 @@
...
+    /// Block is not announced on the network.
+    Private,
+    /// Block is only announced on the local network.
+    Local,
+    /// Block is announced.
+    Public,
+}
+
+/// Implementable by ipld storage backends.
+pub trait Store {
```
aphelionz: Perhaps a base `ReadOnlyStore` that this one `Store: ReadOnlyStore`  for implementations who might just only want `get`?

base/src/store.rs
```diff
@@ -0,0 +1,62 @@
...
+    fn insert(&self, cid: &Cid, data: Box<[u8]>, visibility: Visibility) -> StoreResult<()>;
+
+    /// Flushes the write buffer.
+    fn flush(&self) -> StoreResult<()>;
+
+    /// Marks a block ready for garbage collection.
+    fn unpin(&self, cid: &Cid) -> StoreResult<()>;
+
+    /// Garbage collects unused blocks in the background.
+    fn gc(&self) -> StoreResult<()>;
```
aphelionz: Makes sense but I'm not sure if we want `rust-ipld` to care about pinning or gc? Those two probably deserve a second glance to see if they might actually be better off living in an `ipfs-****` implementation somehow

base/src/store.rs
dvc94ch: In a single user store I'm not sure it makes sense to `pin` a block. Blocks should be pinned on insert and unpinned when they are no longer needed. Unpin is similar to a remove in this case, except that because the store is content addressed, proper liveness tracking needs to happen.

In a multi user store it makes sense. The idea is that you have something like this:
```
user            path
ipfs             /ipfs/block/{cid} 
ipfs             /ipfs/roots/{hash(path)} -> /home/aphelionz/.config/orbitdb/{cid}
aphelionz  /home/aphelionz/.config/orbitdb/{cid} -> /ipfs/block/{cid}
```

aphelionz:
Left a fairly substantial review - thank you again @dvc94ch 

Would love to hear from specifically @vmx + @rklaehn + @mikeal

base/src/store.rs
dvc94ch: Not sure about the use case

aphelionz: For example if i just want to query store passively and not write to i, like a monitoring service or a daily weather report, etc

dvc94ch: Thinking about it maybe it does make sense...

base/src/store.rs
dvc94ch: Unpin is essential for applications. I guess gc is less of an issue as ipld-collections or unixfs probably don't care when to gc.

base/src/ipld.rs
```diff
@@ -21,7 +21,9 @@ pub enum Ipld {
     List(Vec<Ipld>),
     /// Represents a map.
     Map(BTreeMap<String, Ipld>),
-    /// Represents a link to an Ipld node
+    /// Represents a weak link to an Ipld node.
+    WeakLink(Cid),
```
dvc94ch: Based on the proposed semantics for pinning, we need to differentiate between strong and weak links. For backwards compatibility existing links will be weak, and a new encoding added for strong links.

On an ipfs website, the blocks of a page would be a strong link, while a link to a different page might be a weak one. It's up to the application developer to decide.

dvc94ch: A package manager would want to deal in units of packages that may be composed of multiple blocks.

base/src/store.rs
```diff
@@ -0,0 +1,65 @@
...
+
+pub type StoreResult<T> = Pin<Box<dyn Future<Output = Result<T, BlockError>> + Send>>;
+
+/// Visibility of a block.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum Visibility {
+    /// Block is not announced on the network.
+    Private,
+    /// Block is only announced on the local network.
+    Local,
```
dvc94ch: Not sure how useful this option is. It doesn't provide much additional security. Data you want to share within a company network for example should be `Private` and the cid's communicated out-of-band.

dvc94ch: Removed this option to prevent accidental missuse. I think this feature is likely to be missused. Also I think it's an anti-pattern, data is public or it isn't. If the application wants to do this for performance reasons they should configure the store without a dht for example.

base/src/ipld.rs
vmx: Could you please explain a bit more what a weak link is?

dvc94ch: The application determines what it's atoms are. So for example a package manager represents the files of a package as an atomic unit, but the metadata is also ipld and contains a home-page field. That would be a weak link that crosses the application boundaries. Although you're right, that wouldn't be a cid but a name that resolves to a cid. I'm not so sure it does make sense anymore.

dvc94ch: Yep, you convinced me, I don't think this makes sense after all

vmx:
The traits look good to me (without the `WeakLink` change). I just don't want a tight coupling between IPLD and a storage, but these traits actually seem to do exactly that, that you provide building blocks for different storage needs.

base/src/store.rs
aphelionz: I think we can remove `gc` but as I understand you're doing an implementation of this trait currently so I'll wait to see if you end up needing `rust-ipld` to know about it

base/src/store.rs
aphelionz: I'm cool with this as it saves the implementer the trouble of having to include pin if they don't need it

base/src/store.rs
dvc94ch: You're right, I think it's better to remove it and have the implementation schedule gc when it thinks it's best

aphelionz:
Great.

Now... what does this break in `rust-ipfs` and if anything, what's the migration path?

vmx (approved):

dvc94ch:
For one we need to update rust-ipld to actually use it, at the moment implementing the traits doesn't do anything.

rust-ipfs doesn't implement the store trait, but it shouldn't be hard to do. I'll probably build a utility library that helps with some of the details of gc/caching/buffering that I want to share between ipfs-embed and ipld-daemon and remove from rust-ipld.

the other part is ripping the unixfs implementation out of rust-ipfs and make it use rust-ipld. it's also not really in a state that could be called "done".

some notes on graphsync, we should be able to add something like this in the future:
```rust
pub trait SelectorStore {
    fn select(&self, selector: &IpldSelector) -> StoreResult<Vec<Box<[u8]>>>;
}
```

aphelionz (approved):

vmx:
Please make sure next time that Clippy is happy before merging.

Issue #14, opened 2020-02-20
Block API


vmx:
There is a new [Block API in the JavaScript IPLD](https://github.com/ipld/js-block#block-api) implementation, which I find quite nice.

I was thinking about using a similar one for Rust IPLD. I'm still implementing/playing around with the idea, but I thought I publish a draft early on:

Block
------

A `Block` is an IPLD object together with a CID. The data can be encoded
and decoded.

All operations are cached. This means that encoding, decoding and CID
calculation happens at most once. All subsequent calls will use a cached
version.

Methods
-------

#### `pub fn new<R>(cid: Cid, raw: Vec<u8>) -> Self where R: Registry`

Create a new `Block` from the given CID and raw binary data.

It needs a registry that contains codec and hash algorithms
implementations in order to be able to decode the data into IPLD.

#### `pub fn encoder(node: Ipld, codec: &'a dyn Codec<Error = Error>, hash_alg: &'a dyn Hash) -> Self`

Create a new `Block` from the given IPLD object, codec and hash
algorithm.

No computation is done, the CID creation and the encoding will only be
performed when the corresponding methods are called.

#### `pub fn decoder(aw: Vec<u8>, codec: &'a dyn Codec<Error = Error>, hash_alg: &'a dyn Hash) -> Self`

Create a new `Block` from encoded data, codec and hash algorithm.

No computation is done, the CID creation and the decoding will only be
performed when the corresponding methods are called.

#### `pub fn decode(&mut self) -> Ipld`

Decode the `Block` into an IPLD object.

The actual decoding is only performed if the object doesn't have a copy
of the IPLD object yet. If that method was called before, it returns the
cached result.

#### `pub fn encode(&mut self) -> Vec<u8>`

Encode the `Block` into raw binary data.

The actual encoding is only performed if the object doesn't have a copy
of the encoded raw binary data yet. If that method was called before, it
returns the cached result.

#### `pub fn cid(&mut self) -> Cid`

Calculate the CID of the `Block`.

The CID is calculated from the encoded data. If it wasn't encoded
before, that operation will be performed. If the encoded data is already
available, from a previous call of `encode()` or because the `Block` was
instantiated via `encoder()`, then it isn't re-encoded.

dvc94ch:
it seems like you want to have a struct like this:
```rust
struct Block {
    cid: Cid,
    raw: Vec<u8>,
    ipld: Option<Ipld>,
}
```
[...]

PR #33, opened 2020-05-02
Refactor codec traits.

### LINKED ISSUE #24: Removing the Hash trait
Would it make sense to remove the `Hash` trait?

https://github.com/ipfs-rust/rust-ipld/blob/e172fbc8b7c83c751a82c0d4dde877885335152f/base/src/hash.rs#L7-L14

With the new rust-multihash version, it is just the same as `MultihashDigest`:

https://github.com/multiformats/rust-multihash/blob/a7996e06782a703fb983ed665564add0bbf89d51/src/digests.rs#L275-L285

so we could use that directly.

### LINKED ISSUE #27: Replace WriteCbor and ReadCbor.

### LINKED ISSUE #34: add raw codec

dvc94ch:
Refactors the codec trait and updates libipld to use the new store and codec traits. The old store/gc file have been removed.

Depends on https://github.com/multiformats/rust-cid/pull/52 and https://github.com/multiformats/rust-multihash/pull/65

closes #24 closes #27 closes #34 

Created a follow-up issue #35

Moved the block builder to https://github.com/dvc94ch/ipld-block-builder


core/src/hash.rs
```diff
@@ -1,113 +0,0 @@
-//! Hash types.
```
dvc94ch, line 1: TODO: move this to multihash

core/src/mem.rs
```diff
@@ -0,0 +1,79 @@
...
+
+    /// Flushes the write buffer.
+    fn flush<'a>(&'a self) -> StoreResult<'a, ()> {
+        Box::pin(async move { Ok(()) })
+    }
+
+    /// Marks a block ready for garbage collection.
+    fn unpin<'a>(&'a self, cid: &'a Cid) -> StoreResult<'a, ()> {
+        Box::pin(async move {
+            self.blocks.write().await.remove(cid);
```
dvc94ch: this is actually wrong

Cargo.toml
```diff
@@ -45,3 +45,7 @@ harness = false
 [[bench]]
 name = "store"
 harness = false
+
+[patch.crates-io]
+cid = { path = "../rust-cid" }
```
aphelionz: note: we'll need to remove these before merging

core/src/mem.rs
```diff
@@ -0,0 +1,79 @@
+use crate::error::StoreError;
```
aphelionz: As I understand this implementation is a few things - reference implementation for a Store, something built-in that rust-ipld can use if nothing else is implementing a store from the outside... is that correct?

Wondering if this would be better considered as a testing util and/or something in the examples folder? But maybe I'm missing the purpose of this.

aphelionz: The above is mostly just a folder structure nitpick, but I think nailing that down will help future implementers / readers if it's categorized properly.

src/lib.rs
```diff
@@ -1,15 +1,15 @@
 //! The `Ipld` crate.
 
-#![deny(missing_docs)]
+//#![deny(missing_docs)]
```
aphelionz: What about changing this to `#![warn(missing_docs)]` or something that doesn't break CI but also doesn't just eschew code-level docs altogether?

I can see that this PR is substantial and I appreciate that not every single function and trait tediously needs a comment like `// Decodes CBOR` but I'd rather err on the side of over-documenting than under-documenting.

aphelionz (changes requested):
Overall looks really good to me. Thanks for writing the new tests and for doing the legwork of changing the semantics of every encoding type. Just a few minor things and I'm happy to approve, but want @vmx at least to weigh in as well.

core/src/mem.rs
dvc94ch: I think it's useful as a reference implementation. Any store implementation should be "equivalent". If you want to operate on ipld in memory it does make sense to have such an implementation. But they are good points.

aphelionz: Agree it's a great reference

dvc94ch: Well it would be if it were correct, but I think I'll leave that for a follow up.

core/src/mem.rs
dvc94ch: will be fixed in a follow up

dvc94ch:
@vmx any comments? I know you want to make it generic over the `Code`, can that be done in a follow up?

core/src/codec.rs
```diff
@@ -1,18 +1,83 @@
...
-use crate::ipld::Ipld;
-use std::error::Error;
+use std::io::{Read, Write};
 
 /// Codec trait.
-pub trait Codec {
+pub trait Codec: Sized {
     /// Codec version.
-    const VERSION: cid::Version;
+    const VERSION: cid::Version = cid::Version::V1;
```
vmx: I'd remove the version. This is part of a CID and not of a Codec.

core/src/ipld.rs
```diff
@@ -16,7 +17,7 @@ pub enum Ipld {
     /// Represents an UTF-8 string.
     String(String),
     /// Represents a sequence of bytes.
-    Bytes(Vec<u8>),
+    Bytes(Box<[u8]>),
```
vmx: I think vectors are way more common than Boxed slices. So I'm not sure about this change.

core/src/ipld.rs
```diff
@@ -53,21 +54,85 @@ impl<'a> From<&'a str> for IpldIndex<'a> {
     }
 }
 
+/// Ipld type.
+#[derive(Debug)]
+pub enum IpldType {
```
vmx: This sounds very similar what is described as `Node` in the [IPLD Library Design docs](https://github.com/ipld/specs/blob/4eb7b233f38864262c24cc91203098f483f21432/design/libraries/nodes-and-kinds.md#node). So perhaps it would make sense to also introduce the concept of a `Node`.

It would also be good to have some more docs about what differentiates this type from `Ipld`.

core/src/store.rs
```diff
@@ -1,10 +1,10 @@
-use crate::error::BlockError;
+use crate::error::StoreError;
```
vmx: I'm not sure if the storage part should really be part of `core`. But as it currently already is, it is outside of scope for this PR and we should discuss it on a separate issue.

src/block.rs
```diff
@@ -1,88 +1,198 @@
 //! Block validation
-use crate::cid::{Cid, Codec as RawCodec};
-use crate::error::{BlockError, IpldError};
-use crate::hash::{digest, Hash};
+use crate::cid::Cid;
```
vmx, line 2: Here is where my proposal of the Block API (https://github.com/ipfs-rust/rust-ipld/issues/14) comes in. I've created a small demo application which used the proposed API (some details changed, but the idea is the same). The full implementation of that API can be found here: https://github.com/vmx/rust-dagbuilder/blob/f7f0c00b96074f6aa6a5aeef0b327b35d73dea72/src/block.rs

More information is available on the [Block API issue](https://github.com/ipfs-rust/rust-ipld/issues/14#issuecomment-624324812).

I hope that we can use my proposed API (of course after further discussions) instead.

My code currently uses a modified libipld, tomorrow I'll try to base it on this PR, then I can also say more about the encoder/decoder changes.

vmx (commented):
I think we should have a discussion about whether we want to use `Vec<u8>` or `Box<[u8]>` for a lot of things. I need to do more research on it as I currently don't see benefits of using a Boxed slice instead of a vector. So if anyone has pros/cons for either of those, it would be really helpful to post them here.

core/src/codec.rs
dvc94ch: good point

core/src/ipld.rs
dvc94ch: maybe true, but an immutable opaque blob of bytes should be a `Box<[u8]>` not a `Vec<u8>`.

I mentioned in another PR that `Encode` can only be implemented for `Vec<T>` and not for `Vec<T>` and `Vec<u8>`, and same for `Decode`.

core/src/ipld.rs
dvc94ch: This type is only used for constructing `TypeError`.

src/block.rs
dvc94ch: One thing I don't like is this:
```rust
Box::<dyn Codec>::from(self.codec).decode(&raw)
```
There needs to be a trait Code:
```rust
trait Code {
    fn encode(&self) {
        match self {
            Self::Cbor => Cbor::encode(),
        }
    }
    fn decode(&self)
}

impl Code for Codec {}
```
same in multihash there needs to be a trait Code:
```rust
trait Code {
    fn digest()
}
```

dvc94ch: encode and decode should return a reference, the caller can clone if ownership is required.

the block api also doesn't support encoding structs directly, Ipld is required as an intermediate representation.

I don't see necessarily that the two api's are conflicting, they have different use-cases.

dag-cbor/src/encode.rs
```diff
@@ -171,109 +168,118 @@ impl WriteCbor for f64 {
...
+impl Encode<DagCbor> for Cid {
     #[inline]
-    fn write_cbor<W: Write>(&self, w: &mut W) -> Result<()> {
+    fn encode<W: Write>(&self, w: &mut W) -> Result<()> {
         write_tag(w, 42)?;
         // insert zero byte per https://github.com/ipld/specs/blob/master/block-layer/codecs/dag-cbor.md#links
-        let bytes = self.to_bytes();
-        write_u64(w, 2, (bytes.len() + 1) as u64)?;
+        // 128 bytes should be plenty large enough for any cid
+        let mut buf = [0u8; 128];
```
vmx: There is also the identity hash, which is just the data itself. There 128 byte wouldn't be enough.

dvc94ch: the identity hash doesn't have a size restriction?

dvc94ch: I believe that would be a bug. The reason for the identity hash is when the data is smaller than a cryptographic hash function. So it can not be longer than the longest hash function we have which would be 512bits.

https://github.com/multiformats/multihash/issues/13

src/block.rs
vmx: > the block api also doesn't support encoding structs directly, Ipld is required as an intermediate representation.

That's something I haven't thought about yet. But it's a good point. So far I've seen the `Ipld` enum as the central piece everything should center around. But perhaps we can get away with using native types. I need to thing about that.

Also perhaps the block api can be extended to also support native types.

core/src/store.rs
dvc94ch: do you want to open an issue for it?

vmx:
I just want to mention that I had a quick look now and things look good. I'll have a proper review tomorrow, but I wouldn't expect anything major.

Cargo.toml
```diff
@@ -45,3 +45,7 @@ harness = false
 [[bench]]
 name = "store"
 harness = false
+
+[patch.crates-io]
+cid = { git = "https://github.com/multiformats/rust-cid" }
+multihash = { path = "../rust-multihash" }
```
aphelionz: Another reminder to remove this before merging

core/src/convert.rs
```diff
@@ -39,6 +39,7 @@ derive_to_ipld_prim!(Float, f32, clone);
 derive_to_ipld_prim!(Float, f64, clone);
 derive_to_ipld!(String, String, into);
 derive_to_ipld!(String, &str, to_string);
+derive_to_ipld!(Bytes, Box<[u8]>, into_vec);
 derive_to_ipld!(Bytes, Vec<u8>, into);
```
aphelionz, line 43: Looks good. Nice to see both `Box` and `Vec` supported here

dag-json/src/codec.rs
```diff
@@ -7,19 +7,21 @@ use serde_json::ser::Serializer;
...
 use std::fmt;
+use std::io::{Read, Write};
 use std::iter::FromIterator;
 
 const LINK_KEY: &str = "/";
 
-pub fn encode(ipld: &Ipld) -> Result<Box<[u8]>, Error> {
-    let mut writer = Vec::with_capacity(128);
-    let mut ser = Serializer::new(&mut writer);
+#[inline]
```
aphelionz: Why inline only on these? or did I miss others?

aphelionz (changes requested):
Overall looks good, just a few minor things. Also, at some point we'll need a proper README.md with usage and some form of API documentation but that can obviously be left for later.

dag-json/src/codec.rs
dvc94ch: We should probably remove all #[inline] as they increase compile times and recommend that users do lto in release builds.

https://internals.rust-lang.org/t/when-should-i-use-inline/598/2

dvc94ch: Our encode/decode is generic over a Reader/Writer so it may even do inlining without the attribute

core/src/store.rs
vmx: Done: https://github.com/ipfs-rust/rust-ipld/issues/38

dvc94ch: I moved it out of core, is that good enough?

vmx: Yes, sorry, that's good enough. I thought misread `src` for being `core`.

dvc94ch:
One thing left to do is move the errors out of core

koivunej:
If I understand correctly that the changes can be summarized as:

 1. unification of encoding and decoding through respective per codec traits from `WriteCbor` and plain fns for json, pb codecs
 2. internal `ReadCbor` renamed to `TryReadCbor`
 3. inside codecs the errors now are named `Error`
 4. a lot of renames following the (1), (2), (3)
 5. `#[inline]` cleanup
 6. removals in the top level crate

None of these seem major changes in functionality but mostly "streamlining".

Did I miss something?

dvc94ch:
You can call it what you want. But together with the ipld-block-builder and ipfs-embed it provides a complete ipfs implementation.

koivunej:
> You can call it what you want.

I summarized the changes I went through in order for us to make sure I didn't miss anything. I feel like you read me criticizing the work which I did not intend to do.

dvc94ch:
Maintenance is work. The new Codec trait is much more flexible and the status of cbor as a special codec is removed. The only thing special now is that it supports native types, but someone can contribute that to dag-json or dag-pb if they like.

vmx (approved):
Thanks for the effort! Things look good now.

Please wait for one more approval from the Rust IPFS folks, before merging this.

koivunej (approved):

aphelionz (approved):

aphelionz:
Squash one more time and then merge away!

PR #39, opened 2020-05-14
Implement gc for MemStore.

dvc94ch:


src/gc.rs
```diff
@@ -0,0 +1,91 @@
...
+    let mut set: HashSet<Cid> = Default::default();
+    for ipld in ipld.iter() {
+        if let Ipld::Link(cid) = ipld {
+            set.insert(cid.to_owned());
+        }
+    }
+    set
+}
+
+/// Returns the recursive references of an ipld block.
```
vmx, line 21: This might be a non-native speaker (me) thing. I don't associate anything with "recursive references". Reading the code, it looks like you are traversing a DAG starting at some CID and returning all the references that you've visisted.

What about something like:

    /// Traverses a DAG starting at the `roots` and returns all the links it encounters

vmx: A nitpick (something that I would change, but doesn't block merging as it currently is): I find `closure()` a non descriptive name. What about something like `recursive_references()` or `all_references()`.

vmx (commented):
I've only two documentation naming things, rest looks good.

vmx (approved):

PR #42, opened 2020-05-22
Multiblock

### LINKED ISSUE #41: batch insert
A common operation is inserting a batch of blocks and pinning only the last one. This is much more efficient than inserting the blocks individually and then unpinning them. This should be an atomic operation, any leftovers from a failure should be able to be garbage collected.

Required for implementing more sophisticated data structures like vec/hamt, where the changes are computed in memory and then the new blocks inserted in one go.

dvc94ch:
Closes #41


vmx (approved):
Thanks for also renaming the other codecs for consistency and having everything in easily reviewable separate commits.

PR #43, opened 2020-05-28
Allow custom code tables

vmx:
This PR makes `Ipld` generic over code tables. This way you can use IPLD with custom code tables.

For example, by default we should use a specific code table for codecs that are supported by this library instead of using the one bundled with rust-cid. Though, upstream users, e.g. IPFS might want to bundle a different set of codecs, that might even live in external crates. With this change such custom codecs, with their own code table can be used without any modifications of this library.

Not all of the code is fully generic yet. The `MemStore` and parts of `src/block.rs` are missing. Those currently only work with the default codec table.

This PR sets the ground for the Block API I'm working. There are more changes I'd like to make, which follow in future PRs. But I want to keep the PRs self contained, so that issues can easier be discussed.

### DIFF
--- a/dag-cbor/src/decode.rs
+++ b/dag-cbor/src/decode.rs
@@ -439,39 +439,39 @@ where
     fn try_read_cbor<R: Read>(r: &mut R, major: u8) -> Result<Option<Self>> {
         let ipld = match major {
             // Major type 0: an unsigned integer
-            0x00..=0x17 => Ipld::<C, H>::Integer(major as i128),
-            0x18 => Ipld::<C, H>::Integer(read_u8(r)? as i128),
-            0x19 => Ipld::<C, H>::Integer(read_u16(r)? as i128),
-            0x1a => Ipld::<C, H>::Integer(read_u32(r)? as i128),
-            0x1b => Ipld::<C, H>::Integer(read_u64(r)? as i128),
+            0x00..=0x17 => Self::Integer(major as i128),
+            0x18 => Self::Integer(read_u8(r)? as i128),
+            0x19 => Self::Integer(read_u16(r)? as i128),
+            0x1a => Self::Integer(read_u32(r)? as i128),
+            0x1b => Self::Integer(read_u64(r)? as i128),
 
             // Major type 1: a negative integer
-            0x20..=0x37 => Ipld::<C, H>::Integer(-1 - (major - 0x20) as i128),
-            0x38 => Ipld::<C, H>::Integer(-1 - read_u8(r)? as i128),
-            0x39 => Ipld::<C, H>::Integer(-1 - read_u16(r)? as i128),
-            0x3a => Ipld::<C, H>::Integer(-1 - read_u32(r)? as i128),
-            0x3b => Ipld::<C, H>::Integer(-1 - read_u64(r)? as i128),
+            0x20..=0x37 => Self::Integer(-1 - (major - 0x20) as i128),
+            0x38 => Self::Integer(-1 - read_u8(r)? as i128),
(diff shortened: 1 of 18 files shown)


dag-cbor/src/decode.rs
```diff
@@ -411,156 +431,160 @@ impl<T: TryReadCbor> TryReadCbor for BTreeMap<String, T> {
...
+{
     fn try_read_cbor<R: Read>(r: &mut R, major: u8) -> Result<Option<Self>> {
         let ipld = match major {
             // Major type 0: an unsigned integer
-            0x00..=0x17 => Ipld::Integer(major as i128),
-            0x18 => Ipld::Integer(read_u8(r)? as i128),
-            0x19 => Ipld::Integer(read_u16(r)? as i128),
-            0x1a => Ipld::Integer(read_u32(r)? as i128),
-            0x1b => Ipld::Integer(read_u64(r)? as i128),
+            0x00..=0x17 => Ipld::<C, H>::Integer(major as i128),
```
dvc94ch: Changing it to `Self` is probably better

src/store.rs
```diff
@@ -19,20 +21,28 @@ pub enum Visibility {
...
 /// Implementable by ipld storage providers.
-pub trait ReadonlyStore: Clone {
+pub trait ReadonlyStore<C = Codec, H = MultihashCode>: Clone
+where
+    C: Into<u64> + TryFrom<u64> + Copy,
+    H: Into<u64> + TryFrom<u64> + Copy,
+{
     /// Returns a block from the store. If the block is not in the
     /// store it fetches it from the network and pins the block. This
     /// future should be wrapped in a timeout. Dropping the future
```
dvc94ch, line 31: >  This future should be wrapped in a timeout.

Is no longer true. The future returned by the store should timeout if it makes sense.

dag-cbor/src/encode.rs
```diff
@@ -248,18 +253,22 @@ impl<T: Encode<DagCbor> + 'static> Encode<DagCbor> for BTreeMap<String, T> {
...
-            Ipld::Null => write_null(w),
-            Ipld::Bool(b) => Encode::<DagCbor>::encode(b, w),
-            Ipld::Integer(i) => Encode::<DagCbor>::encode(i, w),
-            Ipld::Float(f) => Encode::<DagCbor>::encode(f, w),
-            Ipld::Bytes(b) => Encode::<DagCbor>::encode(b.as_slice(), w),
-            Ipld::String(s) => Encode::<DagCbor>::encode(s, w),
-            Ipld::List(l) => Encode::<DagCbor>::encode(l, w),
-            Ipld::Map(m) => Encode::<DagCbor>::encode(m, w),
-            Ipld::Link(c) => Encode::<DagCbor>::encode(c, w),
+            Ipld::<C, H>::Null => write_null(w),
```
dvc94ch: Self would be better

dvc94ch (commented):
I don't like `CidGeneric` being leaked everywhere. If you are planning on removing the hash table and the codec table from multihash and cid, `CidGeneric` should be renamed to `Cid`.

dag-cbor/src/decode.rs
vmx: That's waaaay better.

src/store.rs
vmx: That's unrelated to this change => separate PR.

vmx:
I would expect that rust-ipld defines it's own `Cid` type, which then users of rust-ipld would use. rust-cid and rust-multihash will probably still bundle some default table, so that they can be used without much hassle.

dvc94ch:
You want to do that in a follow-up?

dvc94ch (approved):

PR #45, opened 2020-06-03
Custom Codec code table

vmx:
This PR introduces a custom Codec code table (instead of using the one from rust-cid) which contains only the supported codecs.

This PR can be reviewed as individual commits, I bundled them into one PR to make it clearer what this is leading to.

### DIFF
--- a/src/block.rs
+++ b/src/block.rs
@@ -54,7 +54,7 @@ where
     C: Into<u64> + TryFrom<u64> + Copy + PartialEq,
     H: Into<u64> + TryFrom<u64> + Copy,
 {
-    if u64::try_from(codec).unwrap() != O::CODE.into() {
+    if codec != O::CODE {
         return Err(Error::UnsupportedCodec(codec.into()));
     }
     D::decode(&mut data).map_err(|e| Error::CodecError(Box::new(e)))
--- a/src/block.rs
+++ b/src/block.rs
@@ -1,17 +1,11 @@
 //! Block validation
 use crate::cid::CidGeneric;
-use crate::codec::{Cid, Codec, Decode, Encode, IpldCodec};
+use crate::codec::{Codec, Decode, Encode, IpldCodec};
+use crate::encode_decode::EncodeDecodeIpld;
 use crate::error::{Error, Result};
 use crate::ipld::Ipld;
 use crate::multihash::{Code as HCode, MultihashDigest, Multihasher};
-use crate::raw::RawCodec;
 use crate::MAX_BLOCK_SIZE;
-#[cfg(feature = "dag-cbor")]
-use libipld_cbor::DagCborCodec;
-#[cfg(feature = "dag-json")]
-use libipld_json::DagJsonCodec;
-#[cfg(feature = "dag-pb")]
-use libipld_pb::DagPbCodec;
 use std::collections::HashSet;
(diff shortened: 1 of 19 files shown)


src/block.rs
```diff
@@ -85,34 +79,23 @@ where
     raw_decode::<C, H, O, D>(cid.codec(), data)
 }
 
-/// Decode raw ipld.
-///
-/// Useful for nested encodings when for example the data is encrypted.
-pub fn raw_decode_ipld(codec: CCode, data: &[u8]) -> Result<Ipld> {
-    match codec {
```
dvc94ch, line 92: Can we keep this? If the data is encrypted the cid hash won't match, as the cid matches the encrypted data

dvc94ch:
In general I have the feeling that multihash, cid and ipld crates are exploding in complexity, I don't feel too happy about that, I'd prefer keeping things simple. But I guess it's ok.

src/block.rs
vmx: Have you seen the intermediate step before it was removed at https://github.com/ipfs-rust/rust-ipld/pull/45/commits/757f94cf43b8ebc9759f07d8e7976e39a3619d51 ?

The `raw_decode_ipld()` call was reduced to basically `codec.decode(data)`. Hence I thought about removing it, as it is easy enough to call that directly. Would that work for your use case?

vmx (commented):
I agree that the explosion of complexity is unfortunate. That's sadly due to the generic nature of IPLD and the way how I'd like to keep in generic (lessons learnt from previous implementations). Though I hope that users of higher level libraries (and I also count users of libipld into that) hopefully won't encounter much of this complexity in their normal usage.

src/block.rs
dvc94ch: I see, makes sense

dvc94ch:
The other issue is I'm finding it unfortunate that multihash/cid aren't stack allocated, as someone could always Box them. And parsing a Sha2256Digest would know the size exactly, and a `Box<dyn Digest>` could parse any multihash

dvc94ch:
I'll probably also implement a codec based on [parity-scale-codec](https://github.com/paritytech/parity-scale-codec) and [scale-info](https://github.com/paritytech/scale-info) crates. `parity-scale-codec` is more efficient than cbor for native data types and scale-info would allow adding a metadata section that could allow decoding it to an ipld representation.

vmx:
> The other issue is I'm finding it unfortunate that multihash/cid aren't stack allocated, as someone could always Box them. And parsing a Sha2256Digest would know the size exactly, and a `Box<dyn Digest>` could parse any multihash

Friedel hacked something together (based on https://github.com/multiformats/rust-cid/pull/52#discussion_r420293672):

https://play.rust-lang.org/?version=nightly&mode=release&edition=2018&gist=296c1d3f20591440cf67ae65db0534e9

to show were things could go in the future. I still haven't had a closer look. But is this what you are after?

dvc94ch:
Yes, something like that. It's unfortunate that const generics aren't a thing yet, as the proposed solution is basically a workaround.

vmx:
@dvc94ch are there any blockers preventing this PR to get merged?

dvc94ch:
No, go ahead.

PR #46, opened 2020-06-10
feat: Block API

vmx:
This commit adds a new Block API. Encoding/decoding only happens when
it is needed. Once such an expensive operation was done, the result
is cached, so that subsequent calls can just return the result.

---

This PR isn't meant as "this should be merged exactly like this". This is meant as a starting point for further discussions. So far when I was talking about the Block API (https://github.com/ipfs-rust/rust-ipld/issues/14), things were pretty abstract, I hope that with this PR things become more concrete. This should illustrate what I'm after. Though getting there could also look completely different.

What I'd like to see that those `Block`s become a integral part, which then e.g. also the storage traits are using.

I know that @dvc94ch use cases are different from mine, so I hope we can find
solutions to both our use cases.

---

One more specific note, I'm not sure if the caching should be implemented for `BlockRef`, it might be nicer/work good enough without it.

### DIFF
--- a/src/block.rs
+++ b/src/block.rs
@@ -105,7 +105,6 @@ where
         if let Some(node) = &self.node {
             Ok(node.clone())
         } else if let Some(raw) = &self.raw {
-            //let decoded = Box::<dyn DynIpldCodec>::from(Box::new(self.codec))._decode(&raw).unwrap();
             let decoded = self.codec.decode(&raw).unwrap();
             self.node = Some(decoded.clone());
             Ok(decoded)
@@ -122,10 +121,6 @@ where
         if let Some(raw) = &self.raw {
             Ok(raw.clone())
         } else if let Some(node) = &self.node {
-            //let encoded = Box::<dyn Codec>::from(self.codec)
-            //    .encode(&node)
-            //    .unwrap()
-            //    .to_vec();
             let encoded = self.codec.encode(&node).unwrap().to_vec();
             self.raw = Some(encoded.clone());
             Ok(encoded)
@@ -147,7 +142,6 @@ where
         if let Some(cid) = &self.cid {
             Ok(cid.clone())
         } else {
-            // TODO vmx 2020-01-31: should probably be `encodeUnsafe()`
             let hash = Box::<dyn MultihashDigest<H>>::from(self.hash_alg).digest(&self.encode()?);
             let cid = CidGeneric::new_v1(self.codec, hash);
             Ok(cid)
@@ -347,7 +341,6 @@ where
             let hash = if let Some(raw) = &self.raw {
                 hasher.digest(raw)
             } else {
-                // TODO vmx 2020-01-31: should probably be `encodeUnsafe()`
                 hasher.digest(&self.encode()?)
             };
             let cid = CidGeneric::new_v1(self.codec, hash);
--- a/src/block.rs
+++ b/src/block.rs
@@ -3,7 +3,7 @@ use crate::cid::CidGeneric;
 use crate::codec::{Codec, Decode, Encode, IpldCodec};
 use crate::encode_decode::EncodeDecodeIpld;
 use crate::ipld::Ipld;
-use crate::multihash::{Code as HCode, MultihashDigest, Multihasher};
+use crate::multihash::{Code as HCode, DigestFromCode, Multihasher};
 use std::collections::HashSet;
 use std::convert::TryFrom;
 use std::fmt;
@@ -142,12 +142,12 @@ where
     /// isn't re-encoded.
     pub fn cid(&mut self) -> Result<&CidGeneric<C, H>, BlockError>
     where
-        Box<dyn MultihashDigest<H>>: From<H>,
+        H: DigestFromCode,
     {
         if let Some(ref cid) = self.cid {
             Ok(cid)
         } else {
-            let hash = Box::<dyn MultihashDigest<H>>::from(self.hash_alg).digest(&self.encode()?);
+            let hash = self.hash_alg.clone().digest(&self.encode()?);
             self.cid = Some(CidGeneric::new_v1(self.codec, hash));
             Ok(self.cid.as_ref().unwrap())
         }
@@ -158,7 +158,6 @@ impl<C, H> fmt::Debug for BlockGeneric<C, H>
 where
     C: Copy + TryFrom<u64> + Into<u64> + fmt::Debug + EncodeDecodeIpld<H>,
     H: Copy + TryFrom<u64> + Into<u64> + fmt::Debug,
-    H: Into<Box<dyn MultihashDigest<H>>>,
     <C as TryFrom<u64>>::Error: fmt::Debug,
     <H as TryFrom<u64>>::Error: fmt::Debug,
 {
@@ -337,7 +336,7 @@ where
     /// isn't re-encoded.
     pub fn cid(&mut self) -> Result<&CidGeneric<C, H>, BlockError>
     where
-        Box<dyn MultihashDigest<H>>: From<H>,
+        H: DigestFromCode,
     {
         // The block was constructed with a CID
         if let Some(cid) = self.cid {
@@ -349,12 +348,11 @@ where
         }
         // The CID needs to be calculated
         else {
-            let hasher = Box::<dyn MultihashDigest<H>>::from(self.hash_alg);
             // We already have the encoded data, use that one directly
             let hash = if let Some(raw) = &self.raw {
-                hasher.digest(raw)
+                self.hash_alg.digest(raw)
             } else {
-                hasher.digest(&self.encode()?)
+                self.hash_alg.clone().digest(&self.encode()?)
             };
             self.cid_cached = Some(CidGeneric::new_v1(self.codec, hash));
             Ok(self.cid_cached.as_ref().unwrap())
--- a/src/block.rs
+++ b/src/block.rs
@@ -12,16 +12,16 @@ use thiserror::Error;
 /// Block API Error.
 #[derive(Error, Debug)]
 pub enum BlockError {
-    /// TDOO
+    /// Block cannot be decoded.
     #[error("Cannot decode block.")]
     DecodeError(String),
-    /// TDOO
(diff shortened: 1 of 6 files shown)


src/block.rs
```diff
@@ -1,101 +1,359 @@
...
+
+    /// Decode the `Block` into an IPLD object.
+    ///
+    /// The actual decoding is only performed if the object doesn't have a copy of the IPLD
+    /// object yet. If that method was called before, it returns the cached result.
+    pub fn decode(&mut self) -> Result<Ipld<C, H>, BlockError> {
+        if let Some(node) = &self.node {
+            Ok(node.clone())
+        } else if let Some(raw) = &self.raw {
+            //let decoded = Box::<dyn DynIpldCodec>::from(Box::new(self.codec))._decode(&raw).unwrap();
```
dvc94ch: please remove unnecessary comment

src/block.rs
```diff
@@ -1,101 +1,359 @@
...
+
+    /// Encode the `Block` into raw binary data.
+    ///
+    /// The actual encoding is only performed if the object doesn't have a copy of the encoded
+    /// raw binary data yet. If that method was called before, it returns the cached result.
+    pub fn encode(&mut self) -> Result<Vec<u8>, BlockError> {
+        if let Some(raw) = &self.raw {
+            Ok(raw.clone())
+        } else if let Some(node) = &self.node {
+            //let encoded = Box::<dyn Codec>::from(self.codec)
```
dvc94ch: same here

src/block.rs
```diff
@@ -1,101 +1,359 @@
...
+            //    .encode(&node)
+            //    .unwrap()
+            //    .to_vec();
+            let encoded = self.codec.encode(&node).unwrap().to_vec();
```
dvc94ch: why is it safe to unwrap here? I don't think it is

src/block.rs
```diff
@@ -1,101 +1,359 @@
...
+            self.raw = Some(encoded.clone());
+            Ok(encoded)
```
dvc94ch: should return a reference and have the caller clone it if necessary

`Ok(self.raw.as_ref().unwrap())`

src/block.rs
```diff
@@ -1,101 +1,359 @@
...
+    /// call of `encode()` or because the `Block` was instantiated via `encoder()`, then it
+    /// isn't re-encoded.
+    pub fn cid(&mut self) -> Result<CidGeneric<C, H>, BlockError>
+    where
+        Box<dyn MultihashDigest<H>>: From<H>,
+    {
+        if let Some(cid) = &self.cid {
+            Ok(cid.clone())
+        } else {
+            // TODO vmx 2020-01-31: should probably be `encodeUnsafe()`
```
dvc94ch: what's encodeUnsafe?

src/block.rs
```diff
@@ -1,101 +1,359 @@
...
+            let hash = Box::<dyn MultihashDigest<H>>::from(self.hash_alg).digest(&self.encode()?);
```
dvc94ch: unnecessary boxing

src/block.rs
```diff
@@ -1,101 +1,359 @@
...
+            Ok((*node).clone())
+        }
+        // The decoder was already called at least once
+        else if let Some(node_cached) = &self.node_cached {
+            Ok(node_cached.clone())
+        }
+        // The data needs to be decoded
+        else if let Some(raw) = &self.raw {
+            let decoded = self.codec.decode(&raw).unwrap();
+            self.node_cached = Some(decoded.clone());
```
dvc94ch: same here caller should clone

dvc94ch (commented):
There's no way of supporting native types now

src/block.rs
mikeal: In the JS API there is `encode/decode` and `encodeUnsafe/decodeUnsafe` methods because both the binary and parsed representations are cached indefinitely and the `Unsafe` methods return those internal cached objects. In JS, we don’t have good immutability primitives, so all we can do is warn people that if they want to use these faster methods they need to make sure they don’t mutate the return values.

Rust has considerably better immutability features so I’d hope that you could find a way to achieve the same results without the `Unsafe` methods.

If I had better immutable features in JS the Block API would have returned immutable objects in both `encode` and `decode` and if someone wanted to mutate them they’d need to copy them first.

vmx:
> There's no way of supporting native types now

I've put thought into this, but couldn't find a good way incorporating that into this Block API. Do you have ideas on this?

I know I've removed those (for clarity), but what about adding functions that do the native conversion? I'd expect that if you want to do native type conversion, you are probably on a lower level and wouldn't want to use the Block API anyway. It would be pure conversions from a native type into a codec an back. E.g. I wouldn't expect higher level things like [IPLD Selectors](https://github.com/ipld/specs/blob/81dcd4caaa472fdf378791ffdee8065ed309fb8d/selectors/selectors.md) would work on them. Or am I missing something?

I currently don't have the native type use case, but you have, so it certainly should be supported.

dvc94ch:
For ipld selectors it would be decoded to ipld. Native types are used when you know the structure, for example AMT, HAMT or unixfsv2 have a very clearly defined structure. If you can't parse it into that structure the data is corrupted. Note that versioning and backwards for native types works by using an enum:

```rust
enum AmtNode {
    V0(AmtNodeV0),
    V1(AmtNodeV1),
}

impl From<AmtNode> for AmtNodeV1 {}
```

src/block.rs
vmx: Thanks for mentioning the `as_ref()` stuff, that saved me a lot of time figuring things out.

src/block.rs
vmx: Hmm, I don't see how I can prevent that. Can you please help?

vmx (commented):
Returning references was a great idea. I think I've addressed everything except for the Multihashing boxing (as I don't know how to do that) and we also need to look into how to support the conversion of native types.

src/block.rs
dvc94ch: I guess it's not really possible now. You'd need to define a trait with the method `digest`, bound H to implement digest, as I've mentioned before.

src/block.rs
```diff
@@ -1,101 +1,368 @@
...
 
-/// Decode block to ipld.
-pub fn decode_ipld<C, H>(cid: &CidGeneric<C, H>, data: &[u8]) -> Result<Ipld<C, H>>
+impl<'a, C, H> BlockRefGeneric<'a, C, H>
 where
-    C: Into<u64> + TryFrom<u64> + Copy + PartialEq + EncodeDecodeIpld<H>,
-    H: Into<u64> + TryFrom<u64> + Copy + PartialEq,
-    Box<dyn MultihashDigest<H>>: From<H>,
+    C: Copy + TryFrom<u64> + Into<u64> + EncodeDecodeIpld<H>,
+    H: Copy + TryFrom<u64> + Into<u64>,
```
dvc94ch, line 222: Note the missing traitbound similar to `EncodeDecodeIpld<H>`

src/block.rs
vmx: The trait for Multihash: https://github.com/multiformats/rust-multihash/pull/68

vmx:
@dvc94ch Could you please have another look? I think I've addressed everything.

I currently point to a branch of rust-multihash, I'll of course get that properly merged and released before I merge this PR.

For encoding and decoding. I tried to keep them both as minimal as possible (e.g. no block size checking). I'd expect that conversion to native types will change a bit, once I get it IPLD Schemas work. I hope that the version I currently propose is good enough for your use case.

dvc94ch:
I'm currently working on a rewrite of ipfs-embed. It'll be tiny and fast. Currently experimenting with a tiny multihash based on dignifiedquires proposal

vmx:
@dvc94ch How does that impact that PR? Is it just about waiting a for the rewrite (which is fine to me), to see if further changes are needed?

src/block.rs
```diff
@@ -1,101 +1,428 @@
...
+            Ok(node)
+        } else if let Some(ref raw) = self.raw {
+            let decoded = self
+                .codec
+                .decode(raw)
+                .map_err(|err| BlockError::DecodeError(err.to_string()))?;
+            self.node = Some(decoded);
+            Ok(self.node.as_ref().unwrap())
+        } else {
+            Err(BlockError::DecodeError(
```
dvc94ch: These are irrecoverable programmer errors. I think an immutable api could be built to make sure these can't happen at compile time.

dvc94ch (commented):
Overall looks good to me for now, but I do think that you could just as well panic.

dvc94ch:
@vmx, I'm not sure yet. ipfs-embed will be targeted at people building blockchain applications with substrate. I'll try to make it as generic as possible, as long as it doesn't add *any* runtime overhead. As a backing db I'm planning to switch to parity-db which is a db optimized and designed for content addressing.

dvc94ch:
@vmx So I have a minimal multihash impl here [0], curious to know what you think. ~Currently it only supports codes up to 127 using `Multihash1`, but I think it would be easy to support up to 16383 for example by adding an enum `Multihash2` where the first variant takes size + 2 bytes to encode and the second one size + 3 bytes. So we'd end up with `Multihash1-9` to fully accomodate the spec. A boxed digest is essentially a `Vec<u8>`.~

That's dumb, by providing your own `Code` enum you know how large it'll be. Problem solved.

- [0] https://github.com/dvc94ch/tiny-multihash

dvc94ch:
I made two drafts #47 #48 . After that we need to see we build an api around it. We probably want a proc macro that will create something like this:
```rust
#[derive(MultiCodec)]
pub enum Code {
    #[mc(code = 0x00, module = crate::cbor::DagCborCodec)]
    DagCbor,
    #[mc(code = 0x01, module = crate::json::DagJsonCodec)]
    DagJson,
}

impl Code {
    fn encode<T: Codec<crate::cbor::DagCborCodec> + Codec<crate::json::DagJsonCodec>>();
    fn decode...
}
```

dvc94ch:
I guess this can be closed for now. If there are still issues, you're welcome to open a new PR.

PR #47, opened 2020-06-19
Use new multihash.

dvc94ch:
This reworks a lot of the outstanding api issues and updates to tiny-mutlihash and tiny-cid.

Still missing a derive proc macro for deriving an IpldCodec. Ipld, Cid, RawMultihash are all weakly typed meaning that the list of supported codecs and multihashes needs to be provided at compile time. This means we can deal with unsupported codecs and hashes, without parse failures. It also means that codecs and hashes are independent of the code, as different tables can map them to different codes.


vmx:
What's the new multihash? One based on https://github.com/multiformats/rust-multihash/pull/69 ?

dvc94ch:
Yes

dvc94ch:
Getting somewhere with it, I think the api could end up something along these lines, there are still a couple issues to work out though. The Wrapper is a hack and since I can't conditionally compile trait bounds it doesn't work. Need to manually implement the matrix of features. This should be less of a problem if crates define their own codecs as they only include what they need.

```rust
type IpldBlock = Block<IpldCodec, Multihash, Wrapper<Ipld>>;

let b1 = IpldBlock::encode(RAW, SHA2_256, &Wrapper(ipld!(&b"cid1"[..]))).unwrap();
let b2 = IpldBlock::encode(DAG_JSON, SHA2_256, &Wrapper(ipld!("cid2"))).unwrap();
let b3 = IpldBlock::encode(DAG_PROTOBUF, SHA2_256, &Wrapper(ipld!("cid3"))).unwrap();

let payload = Wrapper(ipld!({
     "cid1": &b1.cid,
     "cid2": { "other": true, "cid2": { "cid2": &b2.cid }},
     "cid3": [[ &b3.cid, &b1.cid ]],
}));
let block = IpldBlock::encode(DAG_CBOR, SHA2_256, &payload).unwrap();
let payload2 = block.decode().unwrap();
assert_eq!(payload, payload2);

let refs = references(&payload.0);
assert_eq!(refs.len(), 3);
assert!(refs.contains(&b1.cid));
assert!(refs.contains(&b2.cid));
assert!(refs.contains(&b3.cid));
```

vmx:
I only had a brief look, I like what I've seen so far. The multiformats changes seem to help making the code cleaner. Nice work.

The `Encode`/`Decode` trait are taking the codec as a parameter. I wonder why this is needed. If you don't have the time to get into it, that's fine, I guess I'll figure it out once I've a closer look.

dvc94ch:
So only thing missing is updating the DagCbor proc-macro, I'll mark it ready for review

dvc94ch:
Grr still needs work, but I'm getting there. So stuff is getting messy because I'm using the same trait for both a single codec and a codec enum. So we will end up with two types of block.

`Block<Codec, Hasher>` and `Multiblock<Multicodec, Multihash>` with
```rust
impl From<Block<Codec, Hasher>> for Multiblock<Multicodec, Multihash>>
where
    Multicodec: From<Codec>,
    Multihash: From<Hash>;
```
The store traits have a Multicodec and Multihash as associated types and take anything that can be converted in a Multiblock. This is the best we can do api wise as far as I'm concerned.

Also moved the dag traversal into the store trait and added the cache from ipld-block-builder which will be deprecated after this PR is complete.

dvc94ch:
Too complicated. The changes meet from what I can tell all the requirements of sunshine/forest and your block generator cli. I suggest to make this a 0.4 release and be done with it for now.

vmx:
Is it OK, if I have a close look at those changes on Monday (ideally more time, but I don't want to block you) and give feedback before merging?

dvc94ch:
I think it's better than what we currently have in master, although it can certainly be improved. I validated it by porting ipfs-embed and ipld-collections is in the process of being updated by @troublescooter 

I'd appreciate your feedback on Monday.

Fyi, there's a PR open on substrate by @expenses that implements the store traits for substrate using libp2p-bitswap

core/src/codec.rs
```diff
@@ -1,97 +1,49 @@
...
-    type Error: std::error::Error + Send + 'static;
-
     /// Encodes an encodable type.
-    fn encode<T: Encode<Self, C> + ?Sized>(obj: &T) -> Result<Box<[u8]>, Self::Error> {
-        let mut buf = Vec::new();
-        obj.encode(&mut buf)?;
+    fn encode<T: Encode<Self> + ?Sized>(&self, obj: &T) -> Result<Box<[u8]>> {
+        let mut buf = Vec::with_capacity(MAX_BLOCK_SIZE);
+        obj.encode(*self, &mut buf)?;
+        if buf.len() > MAX_BLOCK_SIZE {
```
vmx: Can we please remove the `MAX_BLOCK_SIZE` check here. I really have a strong opinion that this, that it should be an application concern and *not* put directly into the encoding/decoding layer of IPLD.

core/src/codec.rs
```diff
@@ -1,97 +1,49 @@
...
+            return Err(BlockTooLarge(buf.len()).into());
+        }
         Ok(buf.into_boxed_slice())
     }
 
     /// Decodes a decodable type.
-    fn decode<T: Decode<Self, C>>(mut bytes: &[u8]) -> Result<T, Self::Error> {
-        T::decode(&mut bytes)
+    fn decode<T: Decode<Self>>(&self, mut bytes: &[u8]) -> Result<T> {
+        if bytes.len() > MAX_BLOCK_SIZE {
```
vmx: Same es for the encoder.

core/src/error.rs
```diff
@@ -1,8 +1,38 @@
...
+
+/// The block wasn't found. The supplied string is a CID.
+#[derive(Debug, Error)]
+#[error("Failed to retrive block {0}.")]
+pub struct BlockNotFound(pub String);
+
+/// The batch was empty.
+#[derive(Debug, Error)]
+#[error("Tried to insert an empty batch.")]
+pub struct EmptyBatch;
```
vmx, line 34: What is the reason to split those into individual errors and not bundling them into a single error enum? To me it seems common pattern for libraries to errors bundled in an enum.

core/src/error.rs
```diff
@@ -1,8 +1,38 @@
 //! `Ipld` error definitions.
 use crate::ipld::{Ipld, IpldIndex};
-use std::convert::TryFrom;
+pub use anyhow::{Error, Result};
```
vmx, line 3: Do we really want to introduce `anyhow` into `libipld-core`? I've seen the pattern of using `thiserror` for libraries and `anyhow` for applications. I can see that ipld is kind of in-between application and library. Though, is there a good reason to use `anyhow`? Would the code be worse without using `anyhow`?

core/src/codec.rs
```diff
@@ -1,97 +1,49 @@
...
 /// Encode trait.
-pub trait Encode<O, C = IpldCodec>
-where
-    O: Codec<C>,
-    C: Copy + TryFrom<u64> + Into<u64>,
-{
+pub trait Encode<C: Codec> {
     /// Encodes into a `impl Write`.
-    fn encode<W: Write>(&self, w: &mut W) -> Result<(), O::Error>;
+    fn encode<W: Write>(&self, c: C, w: &mut W) -> Result<()>;
```
vmx, line 31: Is the reason to use `c: C` here, so that the call is `Encode::encode(codec_impl, buffer)` instead of `Encode::<codec_impl>::encode(buffer)`? Or is there a deeper reason I don't see?

dag-cbor/src/error.rs
```diff
@@ -0,0 +1,37 @@
...
+
+/// Unexpected eof.
+#[derive(Debug, Error)]
+#[error("Unexpected end of file.")]
+pub struct UnexpectedEof;
+
+/// The byte before Cid was not multibase identity prefix.
+#[derive(Debug, Error)]
+#[error("Invalid Cid prefix: {0}")]
+pub struct InvalidCidPrefix(pub u8);
```
vmx, line 37: The same question as for `core` errors: Why separate Error structs and not a single enum?

dag-cbor/tests/roundtrip.rs
```diff
@@ -29,22 +30,22 @@ fn invalid_cid_prefix() {
...
         "a163666f6fd82a582301122031c3d57080d8463a3c63b2923df5a1d40ad7a73eae5a14af584213e5f504ac33";
     let input = hex::decode(input).unwrap();
-    let result: Result<Ipld, _> = DagCborCodec::decode(&input);
+    let result: Result<Ipld, _> = DagCborCodec.decode(&input);
 
-    match result.unwrap_err() {
-        Error::InvalidCidPrefix(1) => {}
-        x => panic!("unexpected error: {:?}", x),
+    let err = result.unwrap_err();
+    if err.downcast_ref::<InvalidCidPrefix>().is_none() {
```
vmx, line 36: This downcast is needed due to `anyhow`, isn't it?

src/store.rs
```diff
@@ -1,97 +1,95 @@
...
-pub trait ReadonlyStore<C = IpldCodec, H = MultihashCode>: Clone
-where
-    C: Into<u64> + TryFrom<u64> + Copy,
-    H: Into<u64> + TryFrom<u64> + Copy,
-{
+pub trait ReadonlyStore: Clone + Send + Sync {
+    /// The multihash type of the store.
+    type Multihash: MultihashDigest;
+    /// The codec type of the store.
+    type Codec: Codec;
```
vmx, line 21: Double checking if I understand this correctly. All codecs (DagCBOR, DagJSON…)  implement the `Codec` trait. `Multicodec`, which is a wraper around all supported codecs, also implements the `Codec` trait. This way you can have a store that either supports one specific codec (e.g. DagCBOR), or multiple ones (though the `Multicodec` (or any other application specific codecs where you implement the `Codec` trait yourself)). Is that correct?

src/block.rs
```diff
@@ -1,140 +1,191 @@
...
-{
+pub struct Block<C, M> {
+    _marker: PhantomData<(C, M)>,
     /// Content identifier.
-    pub cid: CidGeneric<C, H>,
+    pub cid: Cid,
     /// Binary data.
     pub data: Box<[u8]>,
+    /// Visibility.
+    vis: Visibility,
```
vmx, line 18: I'm not sure if this should be part of the block API. To me networking is a completely different concern. Perhaps we can put that into another layer/level? I need to put more thought into this.

src/block.rs
```diff
@@ -1,140 +1,191 @@
...
+    /// Sets the blocks visibility.
+    pub fn set_visibility(&mut self, vis: Visibility) {
+        self.vis = vis;
     }
-    let hash = Box::<dyn MultihashDigest<H>>::from(cid.hash().algorithm()).digest(&data);
-    if hash.as_ref() != cid.hash() {
-        return Err(Error::InvalidHash(hash.to_vec()));
+
+    /// Encode a block.`
+    pub fn encode<T: Encode<C> + ?Sized>(codec: C, hcode: u64, payload: &T) -> Result<Self> {
```
vmx: Is there a way to unify `encode()` and `encode_ipld()`. I haven't looked deep enough, I just wanted to post this question to make sure I don't forget to dig deeper into that. I'm even not sure if unifying them would be a good idea even if it's possible.

src/block.rs
```diff
@@ -1,140 +1,191 @@
...
-    for ipld in ipld.iter() {
-        if let Ipld::Link(cid) = ipld {
-            set.insert(cid.to_owned());
+    /// Decodes to ipld.
+    pub fn decode_ipld(&self) -> Result<Ipld> {
+        verify_cid::<M>(&self.cid, &self.data)?;
+        C::try_from(self.cid.codec())?.decode_ipld(&self.data)
+    }
+
+    /// Convert from block.
```
vmx: It would be nice to have a bit more explanation what to use it for and perhaps even an example.

src/cache.rs
```diff
@@ -0,0 +1,283 @@
+//! Cache
```
vmx, line 1: I currently don't know why caching should belong directly to IPLD, it sounds like an application level kind of thing. Though atm I also don't have strong objections against having it here.

vmx (commented):
I've also posted a few question, which I can find out myself when actually playing with the code (e.g. what happens i I remove anyhow or put errors into a single enum). But I thought I post my feedback now to shorten the feedback loop and then start playing with it.

What is missing from my proposed Block API is the laziness, which I think is a really nice property. In my Block API you can e.g. have a block which contains and ipld object, but it is encoded only, when it needs to be encoded and not right away. But that should be easy to add to the API.

core/src/codec.rs
dvc94ch: It's so that an enum can also implement the codec trait

core/src/error.rs
dvc94ch: I think it would. Using enums for errors seems like an anti pattern to me. After using error enums and things getting messy I started using this pattern in all my codebases. It should also be faster according to @withoutboats, because the happy path isn't penalized. Most errors are fatal errors not handled but only propagated. In the case that an error needs to be handled by an application downcast_ref works great.

core/src/codec.rs
troublescooter: Block size checks cannot be removed without opening up to DoS attacks. There must be some bound on the amount of data needed to validate a data structure incrementally. So the most flexibility one can hope for is a configurable bound.
Given IPLD transmits the information needed for decoding in-band with the CID, it seems keeping in line with this design one would have to attach a length field.

dvc94ch: It's not an application level concern. It's a networking level concern.

core/src/codec.rs
vmx: Which is a smart move, I really like that!

src/block.rs
dvc94ch: This is not unifiable as it would require expressing that Ipld implements Encode<C>. Rust doesn't support trait bounds on concrete types.

src/block.rs
dvc94ch: A Multicodec can only encode/decode to/from ipld for the mentioned reason. So if you want to use more "features" you need to create a codec specific block and convert it to a block that the store supports

core/src/codec.rs
vmx: Yes, there should be a limit. But I think the limit should be set by the application and not by the IPLD library.

vmx: Yes, there should be a limit. But I think the limit should be set by the application and not by the IPLD library.

dvc94ch: I guess the store could enforce it. But as I said you're only creating a compatibility nightmare, multiblock collections should be used in these cases.

src/block.rs
vmx: Double checking if I understand it correctly. I use a `Multicodec` based `Block` and then I want to store it in a storage that is bould to a `DagCbor` based `Block. Then I can use this method to transform it.

core/src/codec.rs
vmx: There are use cases, where you don't care about the p2p part, or you are sure there are no DoS attacks (internal network). I want that IPLD can be used for such cases as well.

src/store.rs
dvc94ch: Yes

src/block.rs
dvc94ch: It's related to the data, I think it's metadata associated with the block

src/block.rs
dvc94ch: If you get a block from a store you'll get a Multicodec/Multihash block. To decide it to a native types you need to convert it to a DagCbor/Multihash block. The cache is a good example

src/store.rs
vmx: That's really neat!

src/block.rs
vmx: The problem here again is our different view points. I see IPLD as a thing than can be used outside the p2p context, even without networking. I prefer having it those things not directly in those low-level structures.

I also can't tell how many applications really deal with visibility. This is not an IPLD concept (should it be one?), so I I'm not sure it belongs here.

So I'd be in favour of putting it somewhere else, but I also don't have strong objections as users can mostly ignore it if they don't care about it.

vmx: Another idea, perhaps it could be generalized into "meta" which is opaque to IPLD and then applications put in there whatever they want?

dvc94ch: The store needs to know about it.

vmx: I'm also not a fan of coupling a store to tight with such concerns, but that's a separate topic. I also haven't put thought into the store part, hence I'm fine with keeping that all for now. We can revisit when I take a closer look at those parts (if that ever happens). I don't want to block you on your developments.

dag-cbor/tests/roundtrip.rs
dvc94ch: Duplicate of error handling comment

dag-cbor/src/error.rs
dvc94ch: Duplicate

dvc94ch:
So issues remaining:
- error handling strategy
- visibility as an attribute in blocks
- moving blocksize handling to networking layer

dvc94ch:
On the visibility issue, I suggest keeping it this way for now and if we find other metadata that we need to associate with blocks we can revisit it 

I'll move the blocksize restriction to ipfs embed.

On the error handling strategy:
The current system could be improved with a mixture of anyhow and thiserror where errors follow the following pattern:
```rust
enum Error {
    HandleableError()
    Other handleable error()
    Fatal(anyhow::Error)
}
```

dvc94ch:
But the current popular method of nesting more and more error types is imo bad and causes a lot of complications for traits.

vmx:
> So issues remaining:
> 
> * error handling strategy

I don't have enough experience doing it the way you propose (having several types instead of a big enum), so I'm happy to try that out. So that's not an issue anymore.

> The current system could be improved with a mixture of anyhow and thiserror where errors follow the following pattern:
>
> ```rust
> enum Error {
>     HandleableError()
>     Other handleable error()
>     Fatal(anyhow::Error)
> }
> ```

That sounds interesting, but I'm really not sure if it's worth it. My gut feeling is that I wouldn't use `anyhow`, but I'm totally fine with keeping it for now and see how it goes. So this is resolved as well.


> * visibility as an attribute in blocks
>
> On the visibility issue, I suggest keeping it this way for now and if we find other metadata that we need to associate with blocks we can revisit it

I'm fine with keeping it like that for now. I agree we can revisit at a later point in time. Resolved.

> * moving blocksize handling to networking layer
>  I'll move the blocksize restriction to ipfs embed.

I really appreciate that! So that one will then also be resolved.

src/mem.rs
```diff
@@ -2,9 +2,10 @@
 use crate::block::Block;
 use crate::cid::Cid;
 use crate::codec::Codec;
-use crate::error::{BlockNotFound, EmptyBatch, Result};
+use crate::error::{BlockNotFound, BlockTooLarge, EmptyBatch, Result};
 use crate::multihash::MultihashDigest;
 use crate::store::{AliasStore, ReadonlyStore, Store, StoreResult};
+use crate::MAX_BLOCK_SIZE;
```
vmx, line 8: I'm still not a fan of having a hard-coded value in this crate, it should be possible to be set from the outside (application). Especially for the mem storage, I can see that you don't want to have this restriction.

vmx (approved):
As mentioned on a side-channel, the `decode_ipld()` isn't really needed. I'd actually prefer having the ` where Ipld: Codec<Self::Codec>` bound when you use the storage trait. It would make it a nice uniform API. Though I'd need to see what this means in practice.

Though the `decode_ipld()` and the size checking in the storage trait shouldn't block a merge of this PR. Those things can also be changed later on (as I expect that we might change things again, once more libraries use that code).

PR #49, opened 2020-06-29
fix: derive std traits on Block struct

troublescooter:


dvc94ch:
Can you please rebase?

dvc94ch (approved):

PR #50, opened 2020-07-17
Make Store generic, to enable downstream impl of Store trait that don't have to serialise

troublescooter:
The main use case for this is having a store implementation that is useful for debugging. The current Store trait forces to serialise, and leads to unhelpful Debug impl.

This can't be merged since it's based of the 0.3.0 release a few commits back, but is meant as RFC.

### DIFF
--- a/src/block.rs
+++ b/src/block.rs
@@ -15,11 +15,11 @@ use libipld_pb::DagPbCodec;
 use std::collections::HashSet;
 
 /// Block
-pub struct Block {
+pub struct Block<T = Box<[u8]>> {
     /// Content identifier.
     pub cid: Cid,
     /// Binary data.
-    pub data: Box<[u8]>,
+    pub data: T,
 }
 
 /// Encode a block.
--- a/src/mem.rs
+++ b/src/mem.rs
@@ -101,19 +101,16 @@ pub struct MemStore {
     aliases: Arc<RwLock<HashMap<Box<[u8]>, Cid>>>,
 }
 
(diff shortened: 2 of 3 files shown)


dvc94ch (approved):
Looks good to me! I'll hold off on merging for now as there are open api discussions with @vmx

vmx:
@troublescooter I have https://github.com/ipfs-rust/rust-ipld/pull/46 brewing for some time. It is currently paused a bit as @dvc94ch and I independently discovered that we would like to change the rust-multihash API. So I concentrated on that.

I'm not sure I fully understand this issue. So the problem with the current `MemStore` implementation is that it needs to implement `EncodeDecodeIpld`, and you would like to store arbitrary bytes?

troublescooter:
For debugging a multiblock collection having a store that stores generic `T` is a lot more useful than a store that transparently serialises `T` into a stream of bytes. Debug printing the serialising store gives an opaque bunch of bytes of whatever is put in. Another benefit of cutting through the concrete types used in ipld is eliminating the noise of displaying the layered structs.

I'm not convinced of my current proposal though, so I'm open to suggestions. Another option would be to have a cache with two levels, one of non-serialised blocks and another of serialised blocks.

dvc94ch:
Aren't you using the Cache for the hamt anyway? That contains the data in unserialized form. In ipld-block-builder 0.4.0 it's a Cache trait, the cache is now called IpldCache. Wouldn't plugging in a different cache implementation do what you want?

PR #54, opened 2020-08-27
fix: remove unnecessary trait bounds

vmx:
I don't know what they are here for, perhaps some development left-over? If they have their use, it would be nice to have tests that enforces them.

### DIFF
--- a/src/block.rs
+++ b/src/block.rs
@@ -71,10 +71,7 @@ impl<C: Codec, M: MultihashDigest> Block<C, M> {
         codec: CE,
         hcode: u64,
         payload: &T,
-    ) -> Result<Self>
-    where
-        CE: Into<C>,
-    {
+    ) -> Result<Self> {
         let data = codec.encode(payload)?;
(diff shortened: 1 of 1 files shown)


dvc94ch:
Those trait bounds are necessary

dvc94ch:
You can't use a Raw codec if the Multicodec doesn't include a Raw variant

vmx:
> You can't use a Raw codec if the Multicodec doesn't include a Raw variant

Let me try to write a test for that.

dvc94ch:
A block's types are based on the codecs hashes a store supports. The From/Into is used to mark compatibility between a codec and the one the block is parameterized over. To be compatible, the code assigned in the store codec table needs to be the same as the one assigned by the codec.

dvc94ch:
Otherwise there's the issue that the codec will produce an incompatible cid, so we define at compile time that they are compatible, we could check and return an error at runtime instead, but I don't like that solution

dvc94ch:
We should add a debug_assert to check that they result in the same codec code

PR #55, opened 2020-08-27
Add debug assert that codecs produce compatible cid's.

dvc94ch:
Closes #54


vmx (approved):
Cargo fmt failed. Thanks this makes things clearer. I think I'll add some documentation in a subsequent PR to make things even more clear.

Issue #36, opened 2020-05-06
Box<[u8]> vs Vec<u8> as bytes representation


dvc94ch:
My take on it:

1. having different implementations of a trait for `Vec<T>` and `Box<[u8]>` is useful
2. an immutable opaque blob of bytes is a `Box<[u8]>`. a `Vec<u8>` represents a mutable buffer or a list of `u8` that may or may not be bytes.

cc @dignifiedquire @vmx @aphelionz @koivunej @rklaehn

aphelionz:
Perhaps a cop-out answer but it is it possible to support both, at least at the `rust-cid` level, that way implementers of higher-level crates can choose their preference.

I personally prefer `Box<[u8]>` as it signals to the reader the fixed size, but due to `to_vec` and `to_boxed_slice` it also feels like "six in one hand / half dozen in the other."

koivunej:
I haven't seen much of `Box<[u8]>` outside of some std internals, like [`std::io::BufReader`](https://doc.rust-lang.org/src/std/io/buffered.rs.html#56). I'd say the `Vec<u8>` as a growable buffer is quite common, like in [`std::io::Read::read_to_end`](https://doc.rust-lang.org/std/io/trait.Read.html#method.read_to_end).

> an immutable opaque blob of bytes is a `Box<[u8]>`

I'd rather read it as hint of fixed length being somehow *important*, nothing about immutability. I would definitely read a comment explaining why the fixed length was important :) For example in `BufReader` the fixed length is important.

> a `Vec<u8>` represents a mutable buffer or a list of `u8`

Yeah the latter could be possible, while I've rarely seen it used in such a way. Probably because std is full of usages of anything-u8 as a some kind of byte buffer or another. I guess I'd prefer to newtype a `Vec<u8>` if it had more of a "list of" small numbers meaning instead of bytes, since the u8 would likely be used as an optimization of some sort.

dvc94ch:
New typing Vec<u8> doesn't help with implementing traits. I still question the utility of vec methods on bytes. The only thing you want to do on bytes is deref them. The only case I can think of would be the special case where the cid happens to be the first byte sequence in a longer byte sequence. But that's why a copy_to_slice was suggested

vmx:
[...]

PR #57, opened 2020-09-03
Debug info store

dvc94ch:
Still wip.

- removes ReadonlyStore + ReadonlyCache as they're mostly a nuissance
- removes visibility stuff, as the store either is local or networked, but not both. a networked store can still provide a local  store implementation for private data.
- adds pin/status/sync to store
- memstore should support some kind of pseudo networking between multiple mem stores
- refactored memstore implementation
- implement and write more store tests


vmx:
Looks like there's again new ideas coming in. Should I already have a look or wait for bit?

dvc94ch:
Still working on it. Currently rereading the chapter on transactions from "designing data intensive applications". Not sure yet if we require full blown transactions, I think we probably do.

dvc94ch:
I give up. I guess the PR is ready for review

vmx (approved):
Sorry for taking so long for looking at it.

It looks like this changes from `Box<[u8]>` to `Vec<u8>`, could you perhaps post at https://github.com/ipfs-rust/rust-ipld/issues/36 on the reasons? It might be interesting for future reference.

I'm still unsure if `cache` belongs directly to this library, but I'm happy to keep it until proven that it shouldn't be here.

I haven't had a deep look at the storage and cache stuff yet. I will at one point, but I don't want to block merging of this PRs if it's about parts I currently don't care about that much.

The simplification of the Block API (and the removal of `visibility` looks good, hence I'm approving this PR.

dvc94ch:
We'll me adding the collections code and unixfsv2 as crates to this repo soon. They use the cache as caching the raw blocks isn't very useful. Caching the decoded payload is better. I'll leave this open until ipfs-embed 2 is done, as maybe some more changes need to be made as my understanding and experience implementing stores grows

vmx:
> We'll me adding the collections code and unixfsv2 as crates to this repo soon

Do you mean repo or org? I'd prefer if something like Unixfsv2 is something that is cleanly build on top of libipld and isn't directly connect (just using libipld as a depenencey) so I'd like to see it as a separate repo (at least as a separate crate outside of the libipld namespace).

dvc94ch:
Not sure yet, they'd certainly be different crates. It's mostly to make it easier to make changes and coordinate releases.

vmx:
You know that I fear too tight coupling and too large PRs (I don't even want to think about PRs that span several crates in random unrelated ways :)

dvc94ch:
It's not tight coupling. Api's are better if they meet the requirements of what they're built for. rust-ipld was built to decrease coupling and so far we have a very decoupled architecture.

dvc94ch:
Can you review the PR? I finished the ipfs-embed2 storage engine, still writing some tests, but I think this is fine to merge.

vmx (changes requested):
I don't like the block to be coupled with the storage. Especially the Block abstraction should work without any size limits. I guess you can have implementations that ignore it, but it feels strange to define a size limit than then isn't used. I would rather prefer not setting one.

src/store.rs
```diff
@@ -1,110 +1,251 @@
...
     /// The maximum block size supported by the store.
     const MAX_BLOCK_SIZE: usize;
+}
+
+/// Default store parameters.
+#[derive(Clone)]
+pub struct DefaultStoreParams;
+
+impl StoreParams for DefaultStoreParams {
+    const MAX_BLOCK_SIZE: usize = crate::MAX_BLOCK_SIZE;
```
vmx: As I'm still working on my own Block API I don't want to block this version of the of a Block API. What I care about that on the Block level there is by default no size limit. What I can live with is if the `DefaultStoreParams` set the `MAX_BLOCK_SIZE` to `usize::MAX`.

dvc94ch: The block is constructed for a store. Any practical store will have some limitation. Sure we can set it to usize::MAX.

vmx: I'm not sure if binding a Block to a certain store is the right coupling. I would rather have a Block that works independent of the store. Anyway, as I said, I've a different idea of a Block API, so we'll see how all this develops in the future.

dvc94ch: Can you elaborate on the deficiencies in the current api?

vmx: I currently can't. I switched to concentrate on getting tiny-multihash upstreamed, so I haven't had time to work on the block API. As I don't know when I'll have time, I'm OK with merging the current state (with `usize::MAX`) and have a closer look some time later.

dvc94ch: I'm confused. If there is nothing wrong with the current api, why change it?

vmx: I care that there is no block size limit (or a very high one).

I haven't had time to take a closer look at the overall API, hence I can't propose any other changes or say that it works for me.

vmx: So basically I think the API is fine, but I'd like to take a closer look at some point. The only thing that I immediately saw was the size limit, which I don't like.

vmx (approved):
I haven't had a really in-depth review, but what I've seen looks good (better than the old code), hence approving it.

PR #60, opened 2020-09-21
Redesign store.

dvc94ch:


src/mem.rs
```diff
@@ -1,377 +1,348 @@
...
+type Id = u64;
+type Atime = u64;
 
-    /// Returns the number of times the block is pinned.
-    pub fn pinned(&self) -> u32 {
-        self.pinned
-    }
+struct BlockStore<S: StoreParams> {
+    next_id: Id,
+    blocks: HashMap<Id, Block<S>>,
```
rklaehn, line 18: Not sure how relevant this is, but if you care about performance here I would suggest switching to Fnv HashMap. It is the standard hashmap with the fnv hash algorithm. I did some benchmarks for another thing I am working on, and was surprised how big the difference is.

It has 0 dependencies, and you probably have it somewhere in your tree already.

https://crates.io/crates/fnv

image

src/store.rs
```diff
@@ -31,173 +28,24 @@ impl StoreParams for DefaultStoreParams {
...
-    fn into_iter(self) -> Self::IntoIter {
-        self.ops.into_iter()
-    }
-}
-
-/// Implementable by ipld stores.
+/// Implementable by ipld stores. An ipld store behaves like a cache. It will keep blocks
+/// until the cache is full after which it evicts blocks based on an eviction policy. If
+/// a block is aliases, it and it's recursive references will not be evicted or counted towards
+/// the cache size.
```
rklaehn: I would add here that an alias is a named pin, since the concept of pinning is familiar to people coming from go-ipfs.

src/store.rs
```diff
@@ -216,53 +64,49 @@ pub trait Store: Clone + Send + Sync {
...
-            }
-            tx.import(block)?;
-            visited.insert(cid);
-        }
-        tx.update(old, new);
-        self.commit(tx).await
-    }
+    /// Creates an alias for a `Cid`. To alias a block all it's recursive references
+    /// must be in the store. If blocks are missing, they will be fetched from the network. If
+    /// they aren't found, it will return a `BlockNotFound` error.
```
rklaehn, line 69: If the blocks are fetched over the network, a call to this method could take a long time. So when exactly do you return the `BlockNotFound` error? After some configurable timeout?

rklaehn: Coming up with good defaults/heuristics for this is going to be hard. E.g. you create an alias to a single block without links that is just not there, stopping after 2 minutes might be fine. You create an alias to some huge tree, it will take an hour of network traffic to get all the data. So ideally you would stop this after some period of "no progress".

Nevertheless, fine for now. I just anticipate that this part is going to get more complex.

dvc94ch: every fetch has is bounded. in ipfs-embed you can configure it, currently I'm using a 3s by default. so on the first block that times out the operation gets aborted

src/mem.rs
dvc94ch: Interesting. The memstore is really intended for testing or as a reference for people implementing a store. but since it's such a small dependency I made the change.

src/prelude.rs
```diff
@@ -1,4 +1,4 @@
 //! Prelude
-pub use crate::cache::Cache;
+//pub use crate::cache::Cache;
```
vmx: Looks like a development left-over, it can be removed.

vmx (approved):
As with previous store trait changes, I haven't had a deep look and haven't put much thought into a store trait. Though the new code looks simpler, which is most often better. Hence I'm approving it.

rklaehn (approved):
I am unable at this time to give this a really thorough review. But at a glance, looks very good.

src/store.rs
```diff
@@ -216,53 +64,49 @@ pub trait Store: Clone + Send + Sync {
...
+macro_rules! alias {
+    ($name:ident) => {
+        concat!(module_path!(), "::", stringify!($name))
+    };
+}
 
-    /// Removes an alias for a `Cid`.
-    async fn unalias(&self, alias: &[u8]) -> Result<()>;
+/// Creates a dynamic alias by appending a id.
+pub fn dyn_alias(alias: &'static str, id: u64) -> String {
```
troublescooter, line 85: 'static can be removed

troublescooter, line 85: 'static unnecessary.

src/store.rs
```diff
@@ -1,18 +1,15 @@
...
 use crate::multihash::MultihashDigest;
 use crate::path::DagPath;
 use async_trait::async_trait;
-use std::borrow::Cow;
-use std::collections::{HashSet, VecDeque};
-use std::ops::Deref;
 
 /// The store parameters.
-pub trait StoreParams: Clone + Send + Sync {
+pub trait StoreParams: Clone + Send + Sync + Unpin + 'static {
```
troublescooter, line 12: This 'static appears unnecessary, am I missing a reason for it?

dvc94ch: prevents it needing to be propagated. there is no reason why a store params implementation requires a lifetime other than 'static.

src/store.rs
dvc94ch: I think it should be static to ensure it's a compile time string. If you think you know better and want to construct your keys yourself, people are welcome to do so.

PR #61, opened 2020-09-25
Remove unnecessary clone.

troublescooter:


dvc94ch (changes requested):
The logic is wrong, the clone is necessary.

dvc94ch:
Good idea to add some tests:

Can you check that cid/contacts/0/name works? Where the contact is a separate block?

dvc94ch:
I'd be ok adding a take method to ipld to replace the get().clone()

dvc94ch (approved):
You can merge when it's green

src/store.rs
```diff
@@ -19,11 +19,11 @@ pub trait StoreParams: Clone + Send + Sync + Unpin + 'static {
...
 /// Default store parameters.
-#[derive(Clone)]
-pub struct DefaultStoreParams;
+#[derive(Clone, Default)]
+pub struct DefaultParams;
 
-impl StoreParams for DefaultStoreParams {
-    const MAX_BLOCK_SIZE: usize = usize::MAX;
+impl StoreParams for DefaultParams {
+    const MAX_BLOCK_SIZE: usize = 1_048_576;
```
dvc94ch, line 26: @vmx doesn't like this. To some degree it's a parameter of the store, and the MemStore limit is usize

troublescooter:
I shortened DefaultStoreParams to DefaultParams, since the import already specifies what it deals with, and made the default blocksize concrete (1MB), as it seemed unnecessarily undefensive about machine word differences.

src/store.rs
dvc94ch: Can you please revert it? You're aware that it's a contentious issue

troublescooter: I don't see how this change prevents the use case @vmx was going for. A user which does want a different block size limit can create an own struct to impl StoreParams on. I think something called DefaultParams is likely to be used without second thought, and it seems prudent to not support an unreasonable amount of variability in the default case. I certainly didn't expect from cursory reading of the API docs that that limit might differ between 32bit and 64bit machines. Of course, maybe we're a 16bit machine, but at least compilation will tell you about too large literals. Silently compiling a different size however is a bad idea.

dvc94ch: I actually agree and don't object to the change for technical reasons. But you know from previous discussions that @vmx does not, and @vmx and I found a compromise. I don't think sneaking changes that you know are contentious are a good way of collaborating on a project.

troublescooter: I was not aware this was contentious, since I believed the use case of @vmx to be covered. I'll revert the change, but hope to hear from @vmx as to whether it can be made amicably. I should have been less terse about explaining my change and how I thought it covered all bases, waiting for approval.

vmx: @troublescooter thanks for reverting, @dvc94ch thanks for pointing out.

> I think something called DefaultParams is likely to be used without second thought, and it seems prudent to not support an unreasonable amount of variability in the default case.

That's the friction we have. As far as I understand you see libipld in a peer to peer/blockchain context, where such a default limit totally makes sense. I see libipld in a broader context, e.g. I might even store large raw data blocks locally and want to be able to link to them. So for me "default" means "the user can store whatever they like and things don't break".

troublescooter: ~~I'm not sure I can follow. I assumed the StoreParams construction to have been made to accomodate different blocksizes. If your use case is not covered by configuring your store manually, and instead necessitates having all possible stores have a block size limit specified by the machine, what is the configurability of the store for, @dvc94ch ?~~
If you don't provide a manually configured block bound, it's likely that at runtime there will be a crash if a block bigger than `usize::MAX` is deserialised, I could test this to prove it, but I'd rather not. This is unlikely to just work. My proposed fix catches this at compilation instead of runtime.

Edit: I clearly misunderstood the point of contention. Upon rereading, it's not that every block should follow these limits, but that it should be the default, I missed that somehow. The second part of my comment is still relevant though, how would a 32bit machine handle a block larger than machine word size, created on a 64bit machine?

dvc94ch: I'm currently rewriting bitswap to be more robust. The maximum block size will be configurable, so anything below usize::MAX will work. If that's sensible or not is a completely different question. @vmx bitswap will no longer be compatible with pl implementations.

PR #64, opened 2020-09-27
Revert change in default blocksize

troublescooter:


troublescooter:
Reverts the change of default block size introduced in #61

PR #66, opened 2020-09-30
WIP: using new rust-cid

vmx:
This is the WIP to use the (non published yet) rust-cod, which is using multiformats based on the newest tiny-multihash code.

I push it here as I was talking on Matrix about it and I think I found a good solution to having things generic but still usable from the outside.


dvc94ch:
Released 0.7.0

PR #72, opened 2020-12-01
Fast refs.

### LINKED ISSUE #58: Provide mechanism to extract links from a block without building an AST
Building an AST is very wasteful if all you want is a set of links for e.g. pinning.

At least for Dag-CBOR, it is very easy to write a traversal that does not build an AST. See https://gist.github.com/rklaehn/62e21237e7ef8d868cfe655ad6209f28, which uses https://crates.io/crates/cbor_event .

All that would be needed is to extend the codec trait hierarchy for this operation, and then use it, e.g. in https://github.com/ipfs-rust/ipfs-embed/blob/master/src/storage/mod.rs#L75 , to get more efficient traversal of blocks for pinning.

dvc94ch:
Closes #58

### DIFF
--- a/core/src/codec.rs
+++ b/core/src/codec.rs
@@ -1,7 +1,9 @@
 //! `Ipld` codecs.
+use crate::cid::Cid;
 use crate::error::{Result, UnsupportedCodec};
 use core::convert::TryFrom;
-use std::io::{Read, Write};
+use fnv::FnvHashSet;
+use std::io::{Cursor, Read, Seek, Write};
 
 /// Codec trait.
 pub trait Codec:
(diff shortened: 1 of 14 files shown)


core/src/codec.rs
```diff
@@ -44,6 +55,18 @@ pub trait Decode<C: Codec>: Sized {
...
+/// References trait.
+///
+/// This trait is generic over a codec, so that different codecs can be implemented for the same
+/// type.
+pub trait References<C: Codec>: Sized {
+    /// Scrape the references from an `impl Read`.
+    ///
+    /// It takes a specific codec as parameter, so that the [`References`] can be generic over an
+    /// enum that contains multiple codecs.
+    fn references<R: Read + Seek>(c: C, r: &mut R, set: &mut FnvHashSet<Cid>) -> Result<()>;
```
rklaehn: Could we avoid having to have a dependency on fnv by having the caller decide what collection to use for aggregation? Not the most important thing in the world, but some callers might want a different collection.

Maybe leave the set type generic, but just require https://doc.rust-lang.org/std/iter/trait.Extend.html ?

rklaehn (approved):

PR #73, opened 2020-12-01
Disable default hashes.

dvc94ch:
~Waiting on https://github.com/multiformats/rust-multihash/pull/94~


rklaehn (approved):
So this means that you can use libipld core without all the hashes, and when you want hashes you need to specify them in your own multihash dependency? Looks reasonable. I will try it with the importer.

dvc94ch:
Yes that's the idea. You should also be able to use a custom table with exotic hashes.

PR #76, opened 2020-12-16
Repr

### LINKED ISSUE #75: redesign dagcbor macro
a mvp would contain the following representations, making it extensible to add new representations in the future. rust data structures need to be parsed into some kind of intermediate representation before codegen.

enum:
  - string
  - int
union:
  - keyed
  - kinded
struct:
  - map
  - tuple

dvc94ch:
Closes #75 

TODO
- [x] parse structs
- [x] parse unions


vmx:
Wow, nice! I guess it's time that we get test fixtures ready to test IPLD Schema implementations across languages.

dvc94ch:
I commented out the default attribute as it requires more work. The struct fields would likely need to be parsed into an intermediate BTreeMap, so it isn't zero cost.

vmx (approved):
Nice work! In my review I concentrated it on the tests to see if things make sense and they do. Thanks for writing those tests, it made the review way easier.

This also means that I haven't looked super closely at the implementation, but as it works I don't think there are big problems.

PR #89, opened 2021-01-05
New store

dvc94ch:


src/mem.rs
```diff
@@ -1,4 +1,5 @@
 //! Reference store implementation.
+#![allow(warnings)]
```
vmx: Is this a debug left-over?

src/store.rs
```diff
@@ -36,39 +36,55 @@ impl StoreParams for DefaultParams {
 pub trait Store: Clone + Send + Sync {
     /// Store parameters.
     type Params: StoreParams;
+    /// Temp pin.
```
vmx, line 56: Doesn't need to be here, but I think it would be great to have some comment about what a Temp Pin is and what it is used for.

vmx (commented):
I still haven't spend time on thinking about the storage part for libipld, hence I just had a quick look on this PR and only general and no implementation detail level comments.

PR #90, opened 2021-01-21
Changes to ipld data model.

dvc94ch:


vmx:
Could we put those additional things behind a feature flag? I'm not against those, they might have use for some applications that don't have to deal with interoperability with other parts of the IPLD ecosystem. But I'd like also be able to produce data which is spec compatible and hence interoperable with other implementations, as I think this is the power of IPLD.

dvc94ch:
Well to produce compliant data you can not use integer keys or tags. Tags aren't used anywhere anyway, they can only be used if someone implements a custom `Decode/Encode` or manually constructs it using `Ipld::Tag`.

vmx:
> Well to produce compliant data you can not use integer keys or tags

If it's behind a feature flag, you will get errors and not accidentally use them.

vmx:
Thanks for adding the feature flag!

PR #96, opened 2021-02-22
Remove try read cbor

### LINKED ISSUE #91: Remove TryReadCbor trait in favor of seeking
Now that we need to look ahead and need to seek anyway, we should simplify the code and just mandate that when decode fails it hasn't consumed any characters from the read stream.

dvc94ch:
Closes #91

### DIFF
--- a/dag-cbor-derive/src/gen.rs
+++ b/dag-cbor-derive/src/gen.rs
@@ -268,7 +268,6 @@ fn gen_decode_struct(s: &Struct) -> TokenStream {
                         return Ok(#construct);
                     }
                     _ => {
-                        r.seek(SeekFrom::Current(-1))?;
                         return Err(UnexpectedCode::new::<Self>(major).into());
                     }
                 }
@@ -293,7 +292,6 @@ fn gen_decode_struct(s: &Struct) -> TokenStream {
                         return Ok(#construct);
                     }
                     _ => {
-                        r.seek(SeekFrom::Current(-1))?;
                         return Err(UnexpectedCode::new::<Self>(major).into());
                     }
                 }
@@ -316,7 +314,6 @@ fn gen_decode_struct(s: &Struct) -> TokenStream {
                         return Ok(#construct);
                     }
                     _ => {
-                        r.seek(SeekFrom::Current(-1))?;
                         return Err(UnexpectedCode::new::<Self>(major).into());
                     }
                 }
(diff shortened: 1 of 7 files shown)


dag-cbor-derive/src/gen.rs
```diff
@@ -377,15 +350,21 @@ fn gen_try_read_cbor_union(u: &Union) -> TokenStream {
...
-                    let res: Result<Option<Self>> = #parse;
-                    res?;
+                    let result: Result<Self> = (|| {
+                        #parse
+                    })();
+                    match result {
+                        Ok(res) => return Ok(res),
+                        Err(err) if err.downcast_ref::<UnexpectedCode>().is_some() => {}
+                        Err(err) => return Err(err),
+                    };
```
dvc94ch, line 361: Is this acceptable? we will have allocations when a parse tree fails while parsing kinded unions.

rklaehn: For me, yes. That is pretty unlikely to happen on the happy path, unless you have nested kinded unions in the exact right way, right?

rklaehn (approved):
Not sure I can do a meaningful review of this today, but looks good...

dag-cbor/src/decode.rs
```diff
@@ -485,29 +468,25 @@ impl TryReadCbor for Ipld {
...
-                    // TODO this is a hack
-                    r.seek(SeekFrom::Current(-1))?;
-                    return Ok(Some(Self::IntegerMap(read_map(r, len as usize)?)));
+                if len > 0 {
+                    let pos = r.seek(SeekFrom::Current(0))?;
+                    #[cfg(feature = "unleashed")]
+                    if let Ok(map) = read_map(r, len as usize) {
+                        return Ok(Self::IntegerMap(map));
+                    }
+                    r.seek(SeekFrom::Start(pos))?;
```
vmx, line 477: I think this whole if statement could be behind the `unleashed` feature. It is seeking to the current position if `unleashed` isn't enabled.

vmx (commented):
Sorry for being late. I've only one comment, which is just a minor thing.

dvc94ch:
No problem. Don't know how busy you are and what you want to review.

vmx:
Yes, please feel free to go ahead with merging things as you see fit. Don't block on me. I try to review things, when I can, no promises. I guess it's even useful if I would find major things even after a merge.

vmx:
I mostly care that things are reviewed at all, and with @rklaehn helping there it's perfect.

PR #98, opened 2021-02-22
Add SkipOne trait and IgnoredAny and RawValue structs that make use of it

rklaehn:
Not sure about proto, but the SkipOne trait could definitely also be implemented for json. I think proto would work as well...

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -8,6 +8,7 @@ pub mod error;
 pub mod ipld;
 pub mod link;
 pub mod raw;
+pub mod raw_value;
 
 pub use cid;
 pub use multibase;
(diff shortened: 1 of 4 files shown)


rklaehn:
This is pretty simple, and now also has some tests, so I think it is ready to go. Feel free to move things around to where you want them.

vmx (approved):
There's an unused import. but the rest looks good to me.

dvc94ch (approved):
We're acruing a lot of custom cbor walkers. Maybe a visitor pattern isn't that bad after all?

rklaehn:
> We're acruing a lot of custom cbor walkers. Maybe a visitor pattern isn't that bad after all?

We can do that later, since it would not change the public API of IgnoredAny or but tbh I don't mind working with bytes. I will take having to check a byte for a value over wading through a maze of traits like in serde...

PR #100, opened 2021-02-23
Fix indefinite length list

rklaehn:
Allow skipping and reference finding in indefinite length lists


rklaehn:
FYI calling references on the zero length cid example `a163666f6fd82a5800` produces an error. Not sure if that is correct, I just noticed it.

dvc94ch (approved):

dvc94ch:
A zero length CID? what is that?

rklaehn:
No idea. I just copied it from your tests...

PR #101, opened 2021-03-02
Serialize unleashed features.

dvc94ch:
So this fixes the problem of jsonish output for debugging. the entire dag-json thing should be reworked anyway and in my preference not actually read/write compliant json but something that looks like json. But I'm not particulary interested in doing this. Especially since everybody will start complaining again about specs and compatibility.

### DIFF
--- a/dag-json/src/codec.rs
+++ b/dag-json/src/codec.rs
@@ -39,14 +39,15 @@ fn serialize<S: ser::Serializer>(ipld: &Ipld, ser: S) -> Result<S::Ok, S::Error>
             ser.collect_map(wrapped)
         }
         #[cfg(feature = "unleashed")]
-        Ipld::IntegerMap(_) => {
-            // TODO: how to return an error here?
-            unimplemented!()
+        Ipld::IntegerMap(map) => {
+            let wrapped = map.iter().map(|(key, ipld)| (key, Wrapper(ipld)));
+            ser.collect_map(wrapped)
         }
         #[cfg(feature = "unleashed")]
-        Ipld::Tag(_, _) => {
-            // TODO: how to return an error here?
-            unimplemented!()
+        Ipld::Tag(tag, ipld) => {
+            let mut map = BTreeMap::new();
+            map.insert("/", (tag, Wrapper(ipld)));
+            ser.collect_map(map)
         }
         Ipld::Link(link) => {
             let value = base64::encode(&link.to_bytes());


dag-json/src/codec.rs
```diff
@@ -39,14 +39,15 @@ fn serialize<S: ser::Serializer>(ipld: &Ipld, ser: S) -> Result<S::Ok, S::Error>
             ser.collect_map(wrapped)
         }
         #[cfg(feature = "unleashed")]
-        Ipld::IntegerMap(_) => {
-            // TODO: how to return an error here?
-            unimplemented!()
+        Ipld::IntegerMap(map) => {
+            let wrapped = map.iter().map(|(key, ipld)| (key, Wrapper(ipld)));
+            ser.collect_map(wrapped)
```
rklaehn, line 44: So this will produce json like ```{ 1: "foo", 2: "bar" }```?

dag-json/src/codec.rs
```diff
@@ -39,14 +39,15 @@ fn serialize<S: ser::Serializer>(ipld: &Ipld, ser: S) -> Result<S::Ok, S::Error>
...
         }
         #[cfg(feature = "unleashed")]
-        Ipld::Tag(_, _) => {
-            // TODO: how to return an error here?
-            unimplemented!()
+        Ipld::Tag(tag, ipld) => {
+            let mut map = BTreeMap::new();
+            map.insert("/", (tag, Wrapper(ipld)));
+            ser.collect_map(map)
```
rklaehn, line 50: So ```{ "/": (42, "thevalue") }```

rklaehn (approved):
Looks good, but it would be great to have tests that show what the result is.

dag-json/src/codec.rs
dvc94ch: This lets serde_json take care of it, I think this will result in string keys

dag-json/src/codec.rs
dvc94ch: this also lets serde take care of it, I assume it will use an array

dvc94ch:
well, for now it's just for debugging and it's behind the unleashed flag, so the output doesn't really matter.

dag-json/src/codec.rs
rklaehn: serde_json will produce int keys for an int keyed map as far as I know. But that is what I want, so it is fine.

vmx: The keys will likely be a string as JSON doesn't support numbers as keys for objects (only JavaScript does).

rklaehn: Right. Forgot about that. Anyway, this just to have a way to produce debug output for arbitrary dag-cbor unleashed data, so it is fine.

PR #104, opened 2021-03-08
Add support for tuples.

dvc94ch:


dag-cbor/src/decode.rs
```diff
@@ -776,4 +848,14 @@ mod tests {
         let ipld2: Ipld = DagCborCodec.decode(&bytes).unwrap();
         assert_eq!(ipld, ipld2);
     }
+
+    #[test]
+    fn tuples() -> Result<()> {
```
rklaehn, line 853: This does not test all the implemented arities

rklaehn (approved):
Relieved to just see some copy and paste instead of a complex macro!

PR #117, opened 2021-12-06
feat: specialize Vec<u8> (and friends) in cbor encoding

Stebalien:
FOR DISCUSSION, depends on https://github.com/sagebind/castaway/pull/2 anyways.

Previously, `Box<[u8]>` would serialize to a byte string, but `Vec<u8>` would serialize to an array of bytes. Now, they both serialize to a byte string.

The only downside is that Decode is only implemented on over Vec<T> where T is static. But I see no cases where one could reasonably decode into a non-static type.

I believe the new behavior will more closely match the user's expectation, and it'll definitely perform better.

This is all made possible with the `cast` macro from the [castaway](https://github.com/sagebind/castaway/) crate that allows for very basic specialization. That magic _should_ be optimized away at compile time.


dvc94ch:
Interesting, lgtm. maybe time for a changelog as this is a breaking change.

Stebalien:
Awesome! And i'll update the changelog and add some tests.

(meant to submit as a draft)

Stebalien:
> The only downside is that Decode is only implemented on over Vec where T is static. But I see no cases where one could reasonably decode into a non-static type.

Ok, so this is also the case for _encode_, which is unfortunate. I.e., one can't encode `[&someString]`. Is this a common case?

vmx:
While I was working on Serde related things, I stumbled upon what [`serde_bytes`](https://docs.serde.rs/serde_bytes/) is doing. You would annotate the `Vec<u8>`. Perhaps we can also introduce some annotation. This would be more boilerplate code, though less magic, like breaking `[&someString]`.

Or we do the reverse and have some annotation for the `[&someString]` case, which seems way less often used then `Vec<u8>`.

vmx:
I think I'm a fan of having things more explicit and less magic and also similarly to what people are used to (Serde). As we are about to move to Serde for de(serializtion) I don't think there is much point of having this. Hence I'm closing it for now.

Issue #119, opened 2021-12-07
Implement strict DagCbor spec


Stebalien:
See: https://github.com/ipld/ipld/blob/master/specs/codecs/dag-cbor/spec.md#strictness

- no undefined major type.
- no Indefinite length arrays, etc.
- minimal integer encoding.
- f64 only.
- no nan, infinity, etc.

vmx:
I think that issue can be closed, given that the most recent version passed the [IPLD codec fixtures](https://bugzilla.mozilla.org/show_bug.cgi?id=1792370).

PR #120, opened 2021-12-07
feat: stricter cbor

Stebalien:
1. Don't translate undefined to null.
2. Reject indefinite-length items.

Works towards #119

### DIFF
--- a/dag-cbor/src/decode.rs
+++ b/dag-cbor/src/decode.rs
@@ -1,5 +1,8 @@
 //! CBOR decoder
-use crate::error::{InvalidCidPrefix, LengthOutOfRange, UnexpectedCode, UnexpectedEof, UnknownTag};
+use crate::error::{
+    IndefiniteLengthItem, InvalidCidPrefix, LengthOutOfRange, UnexpectedCode, UnexpectedEof,
+    UnknownTag,
+};
 use crate::DagCborCodec as DagCbor;
 use byteorder::{BigEndian, ByteOrder};
 use core::convert::TryFrom;
@@ -84,21 +87,6 @@ pub fn read_list<R: Read + Seek, T: Decode<DagCbor>>(r: &mut R, len: usize) -> R
     Ok(list)
 }
 
-/// Reads a list of any type that implements `TryReadCbor` from a stream of cbor encoded bytes.
-pub fn read_list_il<R: Read + Seek, T: Decode<DagCbor>>(r: &mut R) -> Result<Vec<T>> {
-    let mut list: Vec<T> = Vec::new();
-    loop {
-        let major = read_u8(r)?;
-        if major == 0xff {
-            break;
-        }
-        r.seek(SeekFrom::Current(-1))?;
-        let value = T::decode(DagCbor, r)?;
-        list.push(value);
-    }
-    Ok(list)
-}
-
 /// Reads a map of any type that implements `TryReadCbor` from a stream of cbor encoded bytes.
 pub fn read_map<R: Read + Seek, K: Decode<DagCbor> + Ord, T: Decode<DagCbor>>(
     r: &mut R,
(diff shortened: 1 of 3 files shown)


dvc94ch:
I know that actyx uses some of these features. the indefinite length list was added to be compatible with cbor produced by serde

dvc94ch:
well I guess it doesn't really matter @rkuhn is building his own cbor library for actyx.

Stebalien:
I'm switching over to a next branch for now.

vmx:
> well I guess it doesn't really matter @rkuhn is building his own cbor library for actyx.

Does this mean that we could remove the `unleashed` feature?

@rkuhn: it would be great if you could confirm, whether are removal of that feature flag would be possible.

Background information for those who don't know the `unleashed` feature flag. Actyx needed extended serialization capabilities which are not compatible with IPLD as specified. We decided to introduce a `unleashed` feature that can be enabled if needed. I preferred that over having a fork of the library, only because of such a detail.

rkuhn:
Thanks for asking! My opinion is that using CBOR implies full compliance with RFC8949. If IPLD deviates (for example by removing indefinite-length encoding from the spec) then the result can no longer be called CBOR (and BTW: the PR title is definitely incorrect in this regard).

My opinion on IPLD (which matters a lot less than my opinion on standards as expressed above) is that it should just use plain RFC8949 with no modifications. The IANA registry already has useful tag values defined and the data format is generally a joy to work with. Is there documentation on why the current approach uses something that is not quite CBOR?

BTW: my [CBOR crate](https://crates.io/crates/cbor-data) has a different background as it is intended for dynamic data models, not as a serde replacement.

rkuhn:
I just noticed that I didn’t fully answer your question because I’m not fully aware of what exactly the consequences of removing the `unleashed` feature would be. We store arbitrary CBOR items within IPLD blocks and need to be able to read them as well. These items are not serialised or deserialised using DagCbor, but due to the nature of CBOR the DagCbor deserialiser needs to understand these items so that it can skip over them and hand out the bytes to the layer above.

Useful features of this approach include the ability to use tag 24 (CBOR item) for faster array and dictionary access or tags 0/1 for properly storing timestamps. And we will also make use of the bignum tags (2,3,4,5) inside our payload data model.

vmx:
> the result can no longer be called CBOR

This is the reason why we call it DAG-CBOR. It is a subset of CBOR we use for IPLD. IPLD can have many different codecs so it can't support everything for every codec. It's rather the reverse. We look into the codecs and look how to represent the IPLD Data Model in there. So the basis for IPLD is not CBOR, but the IPLD Data Model.


> I’m not fully aware of what exactly the consequences of removing the `unleashed` feature would be.

It sounds like we should keep the unleash feature then. Though it also means that perhaps the things that got removed with this PR should then be moved into the `unleashed` feature instead? I don't want to break existing users of this library if such a breakage is not strictly needed.

rkuhn:
Ah, thanks for the explanation. To me it then looks like the following:

- DAG-CBOR is a proper subset of CBOR, meaning that any CBOR decoder can be used to read it
- DAG-CBOR is a specific data model on top of CBOR, so encoders need to abide by the DAG-CBOR rules
- this implies that DagCbor is not a suitable serialisation mechanism for the Actyx user payload case, which won’t abide by the rules ⇒ our payload needs to be stored as a normal byte string so that it is skipped by DAG-CBOR

If this is correct then I’d vote for removing the `unleashed` feature. We’ll adapt our code when switching to the upcoming (breaking) version.

vmx:
> To me it then looks like the following:

Yes exactly. This is a good summary.

> If this is correct then I’d vote for removing the `unleashed` feature. We’ll adapt our code when switching to the upcoming (breaking) version.

Perfect. I guess no one else is using the `unleashed` feature. I've opened https://github.com/ipfs-rust/libipld/issues/124 for more visibility and also in case someone else really needs this feature.

PR #123, opened 2021-12-07
fix: avoid recursion when scanning links

Stebalien:


Stebalien:
PRed against next as it would have conflicts on master.

dignifiedquire (commented):
lgtm

PR #125, opened 2021-12-09
fix encoding & decoding

Stebalien:
1. Correctly encode & decode positive signed integers.
2. Correctly handle integer overflow everywhere.
3. Correctly handle all tag types (can be 0-4 bytes).
4. Always encode floats as f64 (by DagCbor spec).
5. Enforce minimal integer encoding.
6. Reject almost everything that's not DagCbor.

To still consider:

1. Be stricter when skipping & scanning for refs:
    1. Validate integers.
    2. Validate floats.
    3. Validate utf8 in strings.
    4. Reject unrecognized tags.
2. Maybe allow on decode:
    1. Undefined.
    2. f16 decoding.
    3. nan, infinity, etc.
3. Maybe forbid on decode:
    1. f16 & f32 on decode.

### DIFF
--- a/dag-cbor-derive/src/gen.rs
+++ b/dag-cbor-derive/src/gen.rs
@@ -18,7 +18,8 @@ pub fn gen_encode(ast: &SchemaType) -> TokenStream {
                 w: &mut W,
             ) -> libipld::Result<()> {
                 use libipld::codec::Encode;
-                use libipld::cbor::encode::{write_null, write_u64};
+                use libipld::cbor::cbor::*;
+                use libipld::cbor::encode::{write_null, write_u8, write_u64};
                 #body
             }
         }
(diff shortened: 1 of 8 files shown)


Stebalien:
@dvc94ch at this point I realize my needs have likely diverged from yours so I'm happy to just fork and move on from there.

dvc94ch (approved):
Lgtm, I don't have utility for this anymore and actyx doesn't care.

vmx (commented):
The changes look good. Once CI is happy it's good to go.

Stebalien:
Merging but I'd still like one more post-merge review just to be sure.

vmx (commented):
Things look good. The only thing I'd like to mention is that you are using glob/star imports for the `cbor` module. I prefer explicit imports as it makes the code easier to grep in. I also find it easier to reason about it while reading, when I know where things come from.

So I suggest changing the `use cbor:*` imports to explicitly listing what gets imported.

PR #129, opened 2022-01-24
Spec compliant DAG-CBOR/DAG-JSON

vmx:
This PR add spec compliant DAG-CBOR/DAG-JSON encoding/decoding. It was locally tested against https://github.com/ipld/codec-fixtures (I'll push that code, once we have a release).

This branch is based on the of the [`next` branch](https://github.com/ipld/libipld/tree/next). The only missing thing were:

 - DAG-CBOR: map key sort order
 - DAG-JSON: supporting the binary kind
 
Here's a link to the diff between `next` branch and this PR: https://github.com/ipld/libipld/compare/next..dag-json-cbor-encoding

### DIFF
--- a/dag-cbor-derive/src/gen.rs
+++ b/dag-cbor-derive/src/gen.rs
@@ -1,3 +1,5 @@
+use std::cmp::Ordering;
+
 use crate::ast::*;
 use proc_macro2::TokenStream;
 use quote::quote;
@@ -108,18 +110,34 @@ fn gen_encode_struct_body(s: &Struct) -> TokenStream {
                     None
                 }
             });
-            let fields = s.fields.iter().map(|field| {
-                let key = rename(&field.name, field.rename.as_ref());
-                let binding = &field.binding;
-                default(
-                    binding,
-                    field.default.as_deref(),
-                    quote! {
-                        Encode::encode(#key, c, w)?;
-                        Encode::encode(#binding, c, w)?;
-                    },
-                )
+            let mut cbor_order = s
+                .fields
+                .iter()
+                .map(|field| {
+                    let key = rename(&field.name, field.rename.as_ref());
+                    let binding = &field.binding;
+                    let field = default(
+                        binding,
+                        field.default.as_deref(),
+                        quote! {
+                            Encode::encode(#key, c, w)?;
+                            Encode::encode(#binding, c, w)?;
+                        },
+                    );
+                    (key.to_string(), field)
+                })
+                .collect::<Vec<(String, _)>>();
+            // CBOR RFC-7049 specifies a canonical sort order, where keys are sorted by length
+            // first. This was later revised with RFC-8949, but we need to stick to the original
+            // order to stay compatible with existing data.
+            cbor_order.sort_unstable_by(|(key_a, _), (key_b, _)| {
+                match key_a.len().cmp(&key_b.len()) {
+                    Ordering::Greater => Ordering::Greater,
+                    Ordering::Less => Ordering::Less,
+                    Ordering::Equal => key_a.cmp(key_b),
+                }
             });
+            let fields = cbor_order.iter().map(|(_, field)| field);
             quote! {
                 let mut len = #len;
                 #(#dfields)*
--- a/dag-cbor-derive/src/gen.rs
+++ b/dag-cbor-derive/src/gen.rs
@@ -20,7 +20,7 @@ pub fn gen_encode(ast: &SchemaType) -> TokenStream {
                 w: &mut W,
             ) -> libipld::Result<()> {
                 use libipld::codec::Encode;
-                use libipld::cbor::cbor::*;
+                use libipld::cbor::cbor::MajorKind;
                 use libipld::cbor::encode::{write_null, write_u8, write_u64};
                 #body
             }
@@ -42,7 +42,7 @@ pub fn gen_decode(ast: &SchemaType) -> TokenStream {
                 c: libipld::cbor::DagCborCodec,
                 r: &mut R,
             ) -> libipld::Result<Self> {
-                use libipld::cbor::cbor::*;
+                use libipld::cbor::cbor::{MajorKind, NULL};
                 use libipld::cbor::decode::{read_uint, read_major};
                 use libipld::cbor::error::{LengthOutOfRange, MissingKey, UnexpectedCode, UnexpectedKey};
                 use libipld::codec::Decode;
--- a/dag-cbor-derive/src/gen.rs
+++ b/dag-cbor-derive/src/gen.rs
@@ -18,7 +18,8 @@ pub fn gen_encode(ast: &SchemaType) -> TokenStream {
                 w: &mut W,
             ) -> libipld::Result<()> {
                 use libipld::codec::Encode;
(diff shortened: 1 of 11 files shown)


vmx:
Here's the WIP branch (pointing to this branch) of the codec fixtures: https://github.com/ipld/codec-fixtures/tree/add-rust

vmx:
I've only rebased to current master. The diff between the `next` branch and this one is still quite small: https://github.com/ipld/libipld/compare/next..dag-json-cbor-encoding

MarcoPolo:
Is the goal to merge this branch into master or next?

vmx:
@MarcoPolo into master.

dag-cbor-derive/src/gen.rs
```diff
@@ -106,22 +110,38 @@ fn gen_encode_struct_body(s: &Struct) -> TokenStream {
...
+                            Encode::encode(#binding, c, w)?;
+                        },
+                    );
+                    (key.to_string(), field)
+                })
+                .collect::<Vec<(String, _)>>();
+            // CBOR RFC-7049 specifies a canonical sort order, where keys are sorted by length
+            // first. This was later revised with RFC-8949, but we need to stick to the original
+            // order to stay compatible with existing data.
+            cbor_order.sort_unstable_by(|(key_a, _), (key_b, _)| {
```
MarcoPolo, line 133: (optional) nit, maybe have a fn called `cbor_rfc_7049_key_sort` with this comment so you don’t have to copy it to multiple places.

dag-json/src/codec.rs
```diff
@@ -168,15 +177,36 @@ impl<'de> de::Visitor<'de> for JsonVisitor {
...
+        if let Some((key, WrapperOwned(Ipld::Map(map)))) = values.first() {
+            // Object with a slash
+            if key == RESERVED_KEY && values.len() == 1 {
+                if let Some((bytes_key, Ipld::String(bytes_value))) = map.iter().next() {
+                    if bytes_key == BYTES_KEY && values.len() == 1 {
+                        let multibase_prefixed = Base::Base64.code().to_string() + bytes_value;
+                        if let Ok((decoded_base, decoded_bytes)) =
+                            multibase::decode(multibase_prefixed)
+                        {
+                            if decoded_base != Base::Base64 {
```
MarcoPolo: How would this happen since you’re prefixing the bytes above?

MarcoPolo: Can we instead use the base64.decode directly and save ourselves from adding this prefix.

maybe something like this, but with error handling:
```
diff --git a/dag-json/src/codec.rs b/dag-json/src/codec.rs
index 16c36bd..490784c 100644
--- a/dag-json/src/codec.rs
+++ b/dag-json/src/codec.rs
@@ -192,15 +192,9 @@ impl<'de> de::Visitor<'de> for JsonVisitor {
             if key == RESERVED_KEY && values.len() == 1 {
                 if let Some((bytes_key, Ipld::String(bytes_value))) = map.iter().next() {
                     if bytes_key == BYTES_KEY && values.len() == 1 {
-                        let multibase_prefixed = Base::Base64.code().to_string() + bytes_value;
-                        if let Ok((decoded_base, decoded_bytes)) =
-                            multibase::decode(multibase_prefixed)
+                        if let Ok(decoded_bytes) = multibase::Base::from_code(Base::Base64.code())
+                            .and_then(|b| b.decode(bytes_value))
                         {
-                            if decoded_base != Base::Base64 {
-                                return Err(SerdeError::custom(
-                                    "bytes kind must be base-64 encoded",
-                                ));
-                            }
                             return Ok(Ipld::Bytes(decoded_bytes));
                         }
                     }
```

MarcoPolo (commented):
Looks good! just one real suggestion around using the base64 decoder directly.

Also it would be cool to see more quickcheck tests. Maybe something that generates an ipld datastructure and round trips it?

dag-cbor-derive/src/gen.rs
vmx: I thought about it, but I didn't want to have more complex dependencies/extra crates only because of this short function. So I decided it's small enough to just have it duplicated.

dag-json/src/codec.rs
vmx: I don't understand your suggestion. The `bytes_value` is a base64 encoded string, without a multibase prefix. That can't be decoded by a multibase implementation.

I'm use multibase and not base64 directly as I didn't want to have another depdendency and also wanted to be sure we use a compatible base64 encoding. I just forgot to remove the `base64` dependency from the crate. Thanks for the pointing me to it, I'll do that now.

vmx:
> Also it would be cool to see more quickcheck tests. Maybe something that generates an ipld datastructure and round trips it?

That sounds like a great addition for a separate PR.

dag-json/src/codec.rs
MarcoPolo: Sorry I wasn't being clear, let me try again:

It seems to me we have this base64 string `bytes_value` and we're trying to decode it. To do that we are currently adding a string prefix to the string, this results in copying `bytes_value` into a new string. I'm saying we don't have to do that and we don't need the base64 crate, we can do it with multibase directly. Since each base in multibase has a `decode` [method](https://docs.rs/multibase/0.9.1/multibase/enum.Base.html#method.decode).

So the diff would look something like this:

```
diff --git a/dag-json/src/codec.rs b/dag-json/src/codec.rs
index 16c36bd..365db1e 100644
--- a/dag-json/src/codec.rs
+++ b/dag-json/src/codec.rs
@@ -192,17 +192,10 @@ impl<'de> de::Visitor<'de> for JsonVisitor {
             if key == RESERVED_KEY && values.len() == 1 {
                 if let Some((bytes_key, Ipld::String(bytes_value))) = map.iter().next() {
                     if bytes_key == BYTES_KEY && values.len() == 1 {
-                        let multibase_prefixed = Base::Base64.code().to_string() + bytes_value;
-                        if let Ok((decoded_base, decoded_bytes)) =
-                            multibase::decode(multibase_prefixed)
-                        {
-                            if decoded_base != Base::Base64 {
-                                return Err(SerdeError::custom(
-                                    "bytes kind must be base-64 encoded",
-                                ));
-                            }
-                            return Ok(Ipld::Bytes(decoded_bytes));
-                        }
+                        let decoded_bytes = Base::Base64
+                            .decode(bytes_value)
+                            .map_err(SerdeError::custom)?;
+                        return Ok(Ipld::Bytes(decoded_bytes));
                     }
                 }
             }
```

Note that `Base::Base64` comes from multibase.

vmx: I'm sorry. I thought `Base::Base64.decode()` would also expect a prefixed string. I didn't look up the API. Thanks for bringing it up, this is indeed a good idea.

MarcoPolo (approved):

vmx:
@MarcoPolo could you please do one last review. I've rebased the PR and added [those two commits](https://github.com/ipld/libipld/pull/129/files/2acf054898dd6b9c9f20dddec71f2743fa0ecbb5..e9c7fdb3023660b85d8b87b6a31d308dc31a433f) where one is kind of based on your review.

MarcoPolo (approved):

PR #133, opened 2022-01-29
dag-pb: update prost to 0.9

dignifiedquire:


vmx:
Just FYI: the current dag-pb implementation doesn't pass the test fixtures at https://github.com/ipld/codec-fixtures.

PR #138, opened 2022-02-07
chore: add cargo-release metadata

vmx:
In order to make releasing easier, add metadata for cargo-release. With this
information you can e.g. run `cargo release --workspace minor` and
it will do the expected thing.

Currently we rely on all crates being bumped to the same version. This is due
to the `shared-version` setting, that makes sure that the commit message has
the `{{version}}` correctly expanded.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,11 @@ license = "MIT OR Apache-2.0"
 description = "library for dealing with ipld"
 repository = "https://github.com/ipld/libipld"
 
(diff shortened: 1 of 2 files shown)


MarcoPolo (approved):
Looks good. For general reference here's the link to what the fields mean: https://github.com/crate-ci/cargo-release/blob/master/docs/reference.md#config-fields (I wasn't familiar with cargo release before)

PR #140, opened 2022-02-15
feat: implement Serde (de)serializer for `Ipld` enum

vmx:
It's now possible to serialize and deserialize the `Ipld` enum with Serde.
As the Serde data model doesn't support custom types, we use the newtype
struct hack, which is used by many other libraries as well. We introduce
a newtype struct with a mostly unique/hard to clash with name. This struct
wraps the CID.

Serde de(serializers) that support CIDs, can then make use of it, others
will see a strange looking struct in their ouput.

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -10,6 +10,7 @@ repository = "https://github.com/ipfs-rust/rust-ipld"
 [features]
 default = ["std"]
 std = ["anyhow/std", "cid/std", "multibase/std", "multihash/std", "thiserror"]
+serde-codec = ["cid/serde-codec", "serde"]
 
 [dependencies]
 anyhow = { version = "1.0.40", default-features = false }
@@ -18,7 +19,10 @@ core2 = { version = "0.4", default-features = false, features = ["alloc"] }
 multihash = { version = "0.16.0", default-features = false, features = ["alloc"] }
 
 multibase = { version = "0.9.1", default-features = false, optional = true }
+serde = { version = "1.0.132", default-features= false, features = ["alloc"], optional = true }
 thiserror = {version = "1.0.25", optional = true }
 
 [dev-dependencies]
 multihash = { version = "0.16.0", default-features = false, features = ["multihash-impl", "blake3"] }
+serde_test = "1.0.132"
+serde_bytes = "0.11.5"
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -19,7 +19,7 @@ core2 = { version = "0.4", default-features = false, features = ["alloc"] }
 multihash = { version = "0.16.0", default-features = false, features = ["alloc"] }
 
 multibase = { version = "0.9.1", default-features = false, optional = true }
-serde = { version = "1.0.132", default-features= false, features = ["alloc"], optional = true }
+serde = { version = "1.0.132", default-features = false, features = ["alloc"], optional = true }
 thiserror = {version = "1.0.25", optional = true }
 
 [dev-dependencies]
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -14,7 +14,7 @@ serde-codec = ["cid/serde-codec", "serde"]
 
 [dependencies]
 anyhow = { version = "1.0.40", default-features = false }
-cid = { version = "0.8.0", default-features = false, features = ["alloc"] }
+cid = { version = "0.8.3", default-features = false, features = ["alloc"] }
 core2 = { version = "0.4", default-features = false, features = ["alloc"] }
 multihash = { version = "0.16.0", default-features = false, features = ["alloc"] }
 
@@ -26,3 +26,4 @@ thiserror = {version = "1.0.25", optional = true }
 multihash = { version = "0.16.0", default-features = false, features = ["multihash-impl", "blake3"] }
 serde_test = "1.0.132"
 serde_bytes = "0.11.5"
+serde_json = "1.0.79"
--- a/core/src/error.rs
+++ b/core/src/error.rs
@@ -36,6 +36,26 @@ pub struct InvalidMultihash(pub Vec<u8>);
 #[cfg_attr(feature = "std", derive(Error), error("Failed to retrieve block {0}."))]
 pub struct BlockNotFound(pub Cid);
 
+/// Error during Serde operations.
+#[cfg(feature = "serde-codec")]
+#[derive(Clone, Debug)]
+#[cfg_attr(feature = "std", derive(Error), error("Serde error: {0}."))]
+pub struct SerdeError(String);
+
+#[cfg(feature = "serde-codec")]
+impl serde::de::Error for SerdeError {
+    fn custom<T: std::fmt::Display>(msg: T) -> Self {
+        Self(msg.to_string())
+    }
+}
+
+#[cfg(feature = "serde-codec")]
+impl serde::ser::Error for SerdeError {
+    fn custom<T: std::fmt::Display>(msg: T) -> Self {
+        Self(msg.to_string())
+    }
+}
+
 /// Type error.
 #[derive(Clone, Debug)]
 #[cfg_attr(
--- a/core/src/error.rs
+++ b/core/src/error.rs
@@ -1,4 +1,6 @@
 //! `Ipld` error definitions.
+#[cfg(feature = "serde-codec")]
+use alloc::string::ToString;
 use alloc::{string::String, vec::Vec};
 
 use crate::cid::Cid;
@@ -39,8 +41,14 @@ pub struct BlockNotFound(pub Cid);
 /// Error during Serde operations.
 #[cfg(feature = "serde-codec")]
 #[derive(Clone, Debug)]
-#[cfg_attr(feature = "std", derive(Error), error("Serde error: {0}."))]
 pub struct SerdeError(String);
+
+#[cfg(feature = "serde-codec")]
+impl core::fmt::Display for SerdeError {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        write!(f, "Serde error: {}", self.0)
+    }
+}
 
 #[cfg(feature = "serde-codec")]
 impl serde::de::Error for SerdeError {
@@ -55,6 +63,9 @@ impl serde::ser::Error for SerdeError {
         Self(msg.to_string())
     }
 }
+
+#[cfg(feature = "serde-codec")]
+impl serde::ser::StdError for SerdeError {}
 
 /// Type error.
 #[derive(Clone, Debug)]
--- a/core/src/error.rs
+++ b/core/src/error.rs
@@ -44,14 +44,14 @@ pub struct SerdeError(String);
 
 #[cfg(feature = "serde-codec")]
 impl serde::de::Error for SerdeError {
-    fn custom<T: std::fmt::Display>(msg: T) -> Self {
+    fn custom<T: core::fmt::Display>(msg: T) -> Self {
         Self(msg.to_string())
     }
 }
 
 #[cfg(feature = "serde-codec")]
 impl serde::ser::Error for SerdeError {
-    fn custom<T: std::fmt::Display>(msg: T) -> Self {
+    fn custom<T: core::fmt::Display>(msg: T) -> Self {
         Self(msg.to_string())
     }
 }
--- a/core/src/ipld.rs
+++ b/core/src/ipld.rs
@@ -7,13 +7,13 @@ use alloc::{
     vec,
     vec::Vec,
 };
-use core::fmt;
+use core::{cmp::Ordering, fmt};
 
 use crate::cid::Cid;
 use crate::error::TypeError;
 
 /// Ipld
-#[derive(Clone, PartialEq)]
+#[derive(Clone, PartialEq, PartialOrd)]
 pub enum Ipld {
     /// Represents the absence of a value or the value undefined.
     Null,
@@ -33,6 +33,21 @@ pub enum Ipld {
     Map(BTreeMap<String, Ipld>),
     /// Represents a map of integers.
     Link(Cid),
+}
+
+// Eq cannot be derived due to the [`Ipld::Float`]. The IPLD Data Model forbids non numeric float
+// values, hence it's OK to implement `Eq`.
+impl Eq for Ipld {}
+
+// TODO vmx 2021-12-21: Make sure that special float values like Nan or Infinity are never
+// constructed and error in case we do.
+/// The sort order is arbitrary.
+#[allow(clippy::derive_ord_xor_partial_ord)]
+impl Ord for Ipld {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.partial_cmp(other)
+            .expect("NaN and Infinity are not supported by IPLD.")
+    }
 }
 
 impl fmt::Debug for Ipld {
--- a/core/src/ipld.rs
+++ b/core/src/ipld.rs
@@ -7,13 +7,13 @@ use alloc::{
     vec,
     vec::Vec,
 };
-use core::{cmp::Ordering, fmt};
+use core::fmt;
 
 use crate::cid::Cid;
 use crate::error::TypeError;
 
 /// Ipld
-#[derive(Clone, PartialEq, PartialOrd)]
+#[derive(Clone, PartialEq)]
 pub enum Ipld {
     /// Represents the absence of a value or the value undefined.
     Null,
@@ -33,21 +33,6 @@ pub enum Ipld {
     Map(BTreeMap<String, Ipld>),
     /// Represents a map of integers.
     Link(Cid),
-}
-
-// Eq cannot be derived due to the [`Ipld::Float`]. The IPLD Data Model forbids non numeric float
-// values, hence it's OK to implement `Eq`.
-impl Eq for Ipld {}
-
-// TODO vmx 2021-12-21: Make sure that special float values like Nan or Infinity are never
-// constructed and error in case we do.
-/// The sort order is arbitrary.
-#[allow(clippy::derive_ord_xor_partial_ord)]
-impl Ord for Ipld {
-    fn cmp(&self, other: &Self) -> Ordering {
-        self.partial_cmp(other)
-            .expect("NaN and Infinity are not supported by IPLD.")
-    }
 }
 
 impl fmt::Debug for Ipld {
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -1,6 +1,6 @@
 //! Core ipld types used by ipld codecs.
 #![deny(missing_docs)]
-#![deny(warnings)]
+//#![deny(warnings)]
 #![cfg_attr(not(feature = "std"), no_std)]
 
 extern crate alloc;
@@ -12,6 +12,8 @@ pub mod ipld;
 pub mod link;
 pub mod raw;
 pub mod raw_value;
+#[cfg(feature = "serde-codec")]
+pub mod serde;
 
 pub use cid;
 #[cfg(feature = "std")]
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -1,6 +1,6 @@
 //! Core ipld types used by ipld codecs.
 #![deny(missing_docs)]
-//#![deny(warnings)]
+#![deny(warnings)]
 #![cfg_attr(not(feature = "std"), no_std)]
 
 extern crate alloc;
--- a/core/src/serde/de.rs
+++ b/core/src/serde/de.rs
@@ -236,7 +236,7 @@ impl<'de> de::Deserializer<'de> for CidDeserializer {
 /// The deserialization will return an error if you try to deserialize into an integer type that
 /// would be too small to hold the value stored in [`Ipld::Integer`].
 ///
-/// [`Ipld::Floats`] are cast, possibly with loss of precision, to the requested Rust type.
+/// [`Ipld::Floats`] can be converted to `f32` if there is no of precision, else it will error.
 impl<'de> de::Deserializer<'de> for Ipld {
     type Error = SerdeError;
 
--- a/core/src/serde/de.rs
+++ b/core/src/serde/de.rs
@@ -1,5 +1,5 @@
-use std::collections::BTreeMap;
-use std::fmt;
+use alloc::{borrow::ToOwned, collections::BTreeMap, format, string::String, vec::Vec};
+use core::fmt;
 
 use cid::serde::{BytesToCidVisitor, CID_SERDE_PRIVATE_IDENTIFIER};
 use cid::Cid;
--- a/core/src/serde/de.rs
+++ b/core/src/serde/de.rs
@@ -288,7 +288,9 @@ impl<'de> de::Deserializer<'de> for Ipld {
                 if !float.is_finite() {
                     error(format!("`Ipld::Float` must be a finite number, not infinity or NaN, input was `{}`", float))
                 } else if (float as f32) as f64 != float {
-                    error("`Ipld::Float` cannot be deserialized to `f32`, without loss of precision`")
+                    error(
+                        "`Ipld::Float` cannot be deserialized to `f32`, without loss of precision`",
+                    )
                 } else {
                     visitor.visit_f32(float as f32)
                 }
@@ -408,10 +410,10 @@ impl<'de> de::Deserializer<'de> for Ipld {
     fn deserialize_tuple_struct<V: de::Visitor<'de>>(
         self,
         _name: &str,
-        _len: usize,
+        len: usize,
         visitor: V,
     ) -> Result<V::Value, Self::Error> {
-        self.deserialize_seq(visitor)
+        self.deserialize_tuple(len, visitor)
     }
 
     fn deserialize_map<V: de::Visitor<'de>>(self, visitor: V) -> Result<V::Value, Self::Error> {
@@ -691,12 +693,18 @@ impl<'de> de::VariantAccess<'de> for VariantDeserializer {
         }
     }
 
-    fn tuple_variant<V>(self, _len: usize, visitor: V) -> Result<V::Value, Self::Error>
+    fn tuple_variant<V>(self, len: usize, visitor: V) -> Result<V::Value, Self::Error>
     where
         V: de::Visitor<'de>,
     {
         match self.0 {
-            Some(Ipld::List(v)) => visit_seq(v, visitor),
+            Some(Ipld::List(list)) => {
+                if len == list.len() {
+                    visit_seq(list, visitor)
+                } else {
+                    error(format!("The tuple variant size must match the length of the `Ipld::List`, tuple variant size: {}, `Ipld::List` length: {}", len, list.len()))
+                }
+            }
             Some(_) => error(format!(
                 "Only `Ipld::List` can be deserialized to tuple variant, input was `{:#?}`",
                 self.0
--- a/core/src/serde/de.rs
+++ b/core/src/serde/de.rs
@@ -285,14 +285,12 @@ impl<'de> de::Deserializer<'de> for Ipld {
     fn deserialize_f32<V: de::Visitor<'de>>(self, visitor: V) -> Result<V::Value, Self::Error> {
         match self {
             Self::Float(float) => {
-                if float.is_finite() {
-                    if (float as f32) as f64 == float {
-                        visitor.visit_f32(float as f32)
-                    } else {
-                        error("`Ipld::Float` cannot be deserialized to `f32`, without loss of precision`")
-                    }
-                } else {
+                if !float.is_finite() {
                     error(format!("`Ipld::Float` must be a finite number, not infinity or NaN, input was `{}`", float))
+                } else if (float as f32) as f64 != float {
+                    error("`Ipld::Float` cannot be deserialized to `f32`, without loss of precision`")
+                } else {
+                    visitor.visit_f32(float as f32)
                 }
             }
             _ => error(format!(
--- a/core/src/serde/de.rs
+++ b/core/src/serde/de.rs
@@ -442,20 +442,11 @@ impl<'de> de::Deserializer<'de> for Ipld {
     fn deserialize_struct<V: de::Visitor<'de>>(
         self,
         _name: &str,
-        fields: &[&str],
+        _fields: &[&str],
         visitor: V,
     ) -> Result<V::Value, Self::Error> {
         match self {
-            Self::Map(map) => {
-                let keys: Vec<_> = map.keys().collect();
-                let mut fields_sorted = fields.to_vec();
-                fields_sorted.sort_unstable();
-                if keys == fields_sorted {
-                    visit_map(map, visitor)
-                } else {
-                    error("Struct fields must match the keys of the `Ipld::Map`")
-                }
-            }
+            Self::Map(map) => visit_map(map, visitor),
             _ => error(format!(
                 "Only `Ipld::Map` can be deserialized to struct, input was `{:#?}`",
                 self
--- a/core/src/serde/de.rs
+++ b/core/src/serde/de.rs
@@ -1,0 +1,713 @@
+use std::collections::BTreeMap;
+use std::fmt;
+
+use cid::serde::{BytesToCidVisitor, CID_SERDE_PRIVATE_IDENTIFIER};
+use cid::Cid;
+use serde::{
+    de::{self, IntoDeserializer},
+    forward_to_deserialize_any,
+};
+
+use crate::error::SerdeError;
+use crate::ipld::Ipld;
+
+/// Deserialize instances of [`crate::ipld::Ipld`].
+///
+/// # Example
+///
+/// ```
+/// use std::collections::BTreeMap;
+///
+/// use serde::Deserialize;
+/// use libipld_core::ipld::Ipld;
+/// use libipld_core::serde::from_ipld;
+///
+/// #[derive(Deserialize)]
+/// struct Person {
+///     name: String,
+///     age: u8,
+///     hobbies: Vec<String>,
+///     is_cool: bool,
+/// }
+///
+/// let ipld = Ipld::Map({
+///     BTreeMap::from([
+///         ("name".into(), Ipld::String("Hello World!".into())),
+///         ("age".into(), Ipld::Integer(52)),
+///         (
+///             "hobbies".into(),
+///             Ipld::List(vec![
+///                 Ipld::String("geography".into()),
+///                 Ipld::String("programming".into()),
+///             ]),
+///         ),
+///         ("is_cool".into(), Ipld::Bool(true)),
+///     ])
+/// });
+///
+/// let person = from_ipld(ipld);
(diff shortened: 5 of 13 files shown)


core/Cargo.toml
```diff
@@ -18,7 +19,10 @@ core2 = { version = "0.4", default-features = false, features = ["alloc"] }
 multihash = { version = "0.16.0", default-features = false, features = ["alloc"] }
 
 multibase = { version = "0.9.1", default-features = false, optional = true }
+serde = { version = "1.0.132", default-features= false, features = ["alloc"], optional = true }
```
Stebalien: ubernit:

```suggestion
serde = { version = "1.0.132", default-features = false, features = ["alloc"], optional = true }
```

core/src/error.rs
```diff
@@ -36,6 +36,26 @@ pub struct InvalidMultihash(pub Vec<u8>);
...
 
+/// Error during Serde operations.
+#[cfg(feature = "serde-codec")]
+#[derive(Clone, Debug)]
+#[cfg_attr(feature = "std", derive(Error), error("Serde error: {0}."))]
+pub struct SerdeError(String);
+
+#[cfg(feature = "serde-codec")]
+impl serde::de::Error for SerdeError {
+    fn custom<T: std::fmt::Display>(msg: T) -> Self {
```
Stebalien: `core::fmt`

Stebalien: It looks like we may need nostd tests.

core/src/ipld.rs
```diff
@@ -35,6 +35,21 @@ pub enum Ipld {
     Link(Cid),
 }
 
+// Eq cannot be derived due to the [`Ipld::Float`]. The IPLD Data Model forbids non numeric float
+// values, hence it's OK to implement `Eq`.
+impl Eq for Ipld {}
+
+// TODO vmx 2021-12-21: Make sure that special float values like Nan or Infinity are never
```
Stebalien: This is also an issue for Eq. Will we have cases that can panic?

core/src/ipld.rs
```diff
@@ -7,13 +7,13 @@ use alloc::{
...
 };
-use core::fmt;
+use core::{cmp::Ordering, fmt};
 
 use crate::cid::Cid;
 use crate::error::TypeError;
 
 /// Ipld
-#[derive(Clone, PartialEq)]
+#[derive(Clone, PartialEq, PartialOrd)]
```
Stebalien: Have we defined ordering somewhere?

- Integer/Float? It should be possible to order between these, but do we want to?
- What about maps? I assume maps simply can't be compared?
- Do we define a "null is less than everything" ordering?

I'd consider punting on this unless we really need it.

Stebalien: I guess this derive will order based on the order in the enum. That's probably not what we want.

core/src/ipld.rs
```diff
@@ -35,6 +35,21 @@ pub enum Ipld {
...
+impl Eq for Ipld {}
+
+// TODO vmx 2021-12-21: Make sure that special float values like Nan or Infinity are never
+// constructed and error in case we do.
+/// The sort order is arbitrary.
+#[allow(clippy::derive_ord_xor_partial_ord)]
+impl Ord for Ipld {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.partial_cmp(other)
+            .expect("NaN and Infinity are not supported by IPLD.")
```
Stebalien: This seems like a pretty big gotcha given that users can construct this value.

core/src/lib.rs
```diff
@@ -1,6 +1,6 @@
 //! Core ipld types used by ipld codecs.
 #![deny(missing_docs)]
-#![deny(warnings)]
+//#![deny(warnings)]
```
Stebalien: delete?

core/src/serde/mod.rs
```diff
@@ -0,0 +1,140 @@
...
+use std::fmt;
+
+use serde::{de::DeserializeOwned, Serialize};
+
+use crate::ipld::Ipld;
+
+/// Utility for testing (de)serialization of [`Ipld`].
+///
+/// Checks if `data` and `ipld` match if they are encoded into each other.
+pub fn assert_roundtrip<T>(data: &T, ipld: &Ipld)
```
Stebalien: Should this really be public if it's for testing?

core/src/serde/ser.rs
```diff
@@ -0,0 +1,454 @@
...
+    value.serialize(Serializer)
+}
+
+impl ser::Serialize for Ipld {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: ser::Serializer,
+    {
+        match &self {
+            Self::Null => serializer.serialize_unit(),
```
Stebalien: Shouldn't this be `serialize_none`?

core/src/serde/ser.rs
```diff
@@ -0,0 +1,454 @@
...
+            Self::Bool(bool) => serializer.serialize_bool(*bool),
+            Self::Integer(i128) => serializer.serialize_i128(*i128),
+            Self::Float(f64) => serializer.serialize_f64(*f64),
```
Stebalien: nit: avoid using type names as variable names, it's a bit confusing.

core/src/serde/ser.rs
```diff
@@ -0,0 +1,454 @@
...
+    }
+
+    #[inline]
+    fn serialize_i8(self, value: i8) -> Result<Self::Ok, Self::Error> {
+        self.serialize_i64(i64::from(value))
+    }
+
+    #[inline]
+    fn serialize_i16(self, value: i16) -> Result<Self::Ok, Self::Error> {
+        self.serialize_i64(i64::from(value))
```
Stebalien, line 128: why not directly to i128?

core/src/serde/ser.rs
```diff
@@ -0,0 +1,454 @@
...
+        Ok(Self::Ok::String(value.to_owned()))
+    }
+
+    fn serialize_bytes(self, value: &[u8]) -> Result<Self::Ok, Self::Error> {
+        Ok(Self::Ok::Bytes(value.to_vec()))
+    }
+
+    #[inline]
+    fn serialize_unit(self) -> Result<Self::Ok, Self::Error> {
+        Ok(Self::Ok::Null)
```
Stebalien: Honestly, I'd serialize this as an empty list, not as null. Unit is `()` (emtpty tuple).

core/src/serde/ser.rs
```diff
@@ -0,0 +1,454 @@
...
+    }
+
+    #[inline]
+    fn serialize_unit_struct(self, _name: &'static str) -> Result<Self::Ok, Self::Error> {
+        self.serialize_unit()
```
Stebalien: ditto above. That or, really we could just _reject_ unit.

core/src/serde/ser.rs
```diff
@@ -0,0 +1,454 @@
...
+    }
+
+    #[inline]
+    fn serialize_unit_variant(
+        self,
+        _name: &'static str,
+        _variant_index: u32,
+        variant: &'static str,
+    ) -> Result<Self::Ok, Self::Error> {
+        self.serialize_str(variant)
```
Stebalien, line 206: Honestly, I'd expect this to use the variant index, at least in rust.

Maybe we should just not implement this? I.e., force the user to specify their representation when they derive serialize on their type?

core/src/serde/ser.rs
Stebalien: Alternatively, it would make sense to serialize this as the name itself (as a constant). But... I'd still prefer to just not implement it.

Stebalien (commented):
Initial review. My main feedback would be to reject more types, forcing the user to specify the representation.

core/src/error.rs
vmx: Currently building `cargo build --no-default-features --features serde-codec --target thumbv6m-none-eabi --manifest-path core/Cargo.toml` fails due to Serde. So that might be difficult to test.

core/src/serde/ser.rs
vmx: That's also what `serde_json` is doing:
https://github.com/serde-rs/json/blob/7025523603fe604d11b92ccd4ca314e343d3ae50/src/value/ser.rs#L22

And for me personally Rust's `None` kind of also implies there's a `Some`, which obviously isn't the case for `null`.

core/src/serde/ser.rs
vmx: It's what `serde_json` is doing: https://github.com/serde-rs/json/blob/7025523603fe604d11b92ccd4ca314e343d3ae50/src/value/ser.rs#L160-L162 To me it makes sense that we do the same.

core/src/serde/ser.rs
vmx: Same again, that's what `serde_json` is doing: https://github.com/serde-rs/json/blob/7025523603fe604d11b92ccd4ca314e343d3ae50/src/value/ser.rs#L165-L167

core/src/serde/ser.rs
vmx: This is again what `serde_json` is doing: https://github.com/serde-rs/json/blob/7025523603fe604d11b92ccd4ca314e343d3ae50/src/value/ser.rs#L165-L167

vmx:
I think at one point I needed `Eq`, `Ord` and `PartialOrd` for something fundamental. I can't recall for what. As things seems to work without it, I'd just not implement them for now.

In regards to "should we serialize to xyz", I think it makes sense to mirror `serde_json`s behaviour as IPLDs data model is so similar and I as a user would also expect it to work in a similar way. WDYT @Stebalien.

All other things should now be fixed.

core/src/error.rs
Stebalien: Hm. Just building with `#![no_std]` should catch this, even without `thumbv6m-none-eabi`.

core/src/serde/ser.rs
Stebalien: Serde is too automagical, IMO. I can't think of why I, as a user, would want unit and/or a unit struct serialized to "null" automatically.

core/src/serde/ser.rs
Stebalien: See below.

core/src/serde/ser.rs
Stebalien: Basically, serde's goal is to "just work". But with IPLD, we really care about the specifics.

core/src/serde/ser.rs
Stebalien: See above. Especially in this case where, personally, I'd expect it to encode the enum as the underlying representation. E.g.:

```rust
#[repr(u8)]
enum Foo {
    Yes = 1,
    No = 2,
}
```

core/src/serde/ser.rs
Stebalien: Actually, let's just have this discussion in the comments.

core/src/serde/ser.rs
Stebalien: Eh, discuss below.

Stebalien:
> In regards to "should we serialize to xyz", I think it makes sense to mirror serde_jsons behaviour as IPLDs data model is so similar and I as a user would also expect it to work in a similar way. WDYT @Stebalien.

I think IPLD is significantly different:

- Serde's attitude is "just do the thing". I.e., as long as it round-trips with serde, it's good enough.
- For us, we very much care about the final representation and about not having any magic.

If we find that not having some specific magic is becoming a problem, we always add more magic as necessary. But it's hard to take it away if we find that it's confusing and/or unexpected.

For example:

- Given a unit type, I wouldn't expect it to encode to whatever the unit type's name is in rust. Really, I'd expect either:
   - The user to _skip_ the field (i.e., it's a bug).
   - The unit type to encode to some constant.
- Given an enum with no "values", I'd expect it to encode to the tag values (1, 2, 3, etc.) or some "key". But, in most cases, I'd expect the user to specify the key.

core/src/serde/ser.rs
vmx: > And for me personally Rust's `None` kind of also implies there's a `Some`, which obviously isn't the case for `null`.

Which actually is the case if you use `Ipld::Null`, so it makes sense representing it as `None`.

vmx:
@Stebalien thanks for explaning. I had another look and came to the same conclusions. Here's the current state:

 - `Ipld::Null` now serializes to `None`
 - unit and unit structs error. They can't neither be serialized, nor deserialized. I think symmetry makes sense here.
 - The only thing I haven't done that you brought up is about enums with e.g. `repr(u8)`. I think representing as enum by default as string makes sense. If you want the actual integer value, you should be able to use [`serde_repr`](https://serde.rs/enum-number.html).

Stebalien:
> The only thing I haven't done that you brought up is about enums with e.g. repr(u8). I think representing as enum by default as string makes sense. If you want the actual integer value, you should be able to use [serde_repr](https://serde.rs/enum-number.html).

Fair enough. We can also revisit the other decisions later if we change our minds.

core/src/serde/de.rs
```diff
@@ -0,0 +1,693 @@
...
+            {
+                Ok(Ipld::Null)
+            }
+
+            #[inline]
+            fn visit_seq<V>(self, mut visitor: V) -> Result<Self::Value, V::Error>
+            where
+                V: de::SeqAccess<'de>,
+            {
+                let mut vec = Vec::new();
```
Stebalien: nit: use `visitor.size_hint()`.

core/src/serde/de.rs
```diff
@@ -0,0 +1,693 @@
...
+                while let Some((key, value)) = visitor.next_entry()? {
+                    values.insert(key, value);
+                }
+
+                Ok(Ipld::Map(values))
+            }
+
+            /// Newtype structs are only used to deserialize CIDs.
+            #[inline]
+            fn visit_newtype_struct<D>(self, deserializer: D) -> Result<Self::Value, D::Error>
```
Stebalien, line 175: We should test this with non dag codecs to make sure we don't get any strange results (e.g, infinite recursion).

core/src/serde/de.rs
```diff
@@ -0,0 +1,693 @@
...
+        deserializer.deserialize_any(IpldVisitor)
+    }
+}
+
+macro_rules! impl_deserialize_integer {
+    ($ty:ident, $deserialize:ident, $visit:ident) => {
+        fn $deserialize<V: de::Visitor<'de>>(self, visitor: V) -> Result<V::Value, Self::Error> {
+            match self {
+                Self::Integer(integer) => {
+                    if integer >= $ty::MIN.into() && integer <= $ty::MAX.into() {
```
Stebalien: consider using `let int: $ty = integer.try_into()?`?

core/src/serde/de.rs
```diff
@@ -0,0 +1,693 @@
...
+/// A separate deserializer is needed to make sure we always deserialize only CIDs as `Ipld::Link`
+/// and don't deserialize arbitrary bytes.
+struct CidDeserializer(Cid);
+
+impl<'de> de::Deserializer<'de> for CidDeserializer {
+    type Error = SerdeError;
+
+    #[inline]
+    fn deserialize_any<V: de::Visitor<'de>>(self, _visitor: V) -> Result<V::Value, Self::Error> {
+        unreachable!()
```
Stebalien: What if we use `Ipld` with, e.g., non-dag json? I think we need to return a type error instead of panicing.

core/src/serde/de.rs
```diff
@@ -0,0 +1,693 @@
...
+        match self {
+            Self::Null => visitor.visit_none(),
+            Self::Bool(bool) => visitor.visit_bool(bool),
+            Self::Integer(i128) => visitor.visit_i128(i128),
+            Self::Float(f64) => visitor.visit_f64(f64),
+            Self::String(string) => visitor.visit_str(&string),
+            Self::Bytes(bytes) => visitor.visit_bytes(&bytes),
+            Self::List(list) => visit_seq(list, visitor),
+            Self::Map(map) => visit_map(map, visitor),
+            Self::Link(cid) => visitor.visit_newtype_struct(CidDeserializer(cid)),
```
Stebalien, line 257: Maybe we should just fail here? I.e., won't this deserialize into any:

```rust
struct NewType(Bytes);
```

?

Stebalien: Hm. Well, we need to support deserializing IPLD -> IPLD...

core/src/serde/de.rs
```diff
@@ -0,0 +1,693 @@
...
+    impl_deserialize_integer!(i64, deserialize_i64, visit_i64);
+
+    impl_deserialize_integer!(u8, deserialize_u8, visit_u8);
+    impl_deserialize_integer!(u16, deserialize_u16, visit_u16);
+    impl_deserialize_integer!(u32, deserialize_u32, visit_u32);
+    impl_deserialize_integer!(u64, deserialize_u64, visit_u64);
+
+    fn deserialize_f32<V: de::Visitor<'de>>(self, visitor: V) -> Result<V::Value, Self::Error> {
+        match self {
+            Self::Float(float) => visitor.visit_f32(float as f32),
```
Stebalien: Hm. Don't we want some form of "fits" test?

core/src/serde/de.rs
```diff
@@ -0,0 +1,693 @@
...
+                "Only `Ipld::Float` can be deserialized to `f64`, input was `{:#?}`",
+                self
+            )),
+        }
+    }
+
+    fn deserialize_char<V: de::Visitor<'de>>(self, visitor: V) -> Result<V::Value, Self::Error> {
+        match self {
+            Self::String(string) => {
+                if string.len() == 1 {
```
Stebalien: Length is bytes (https://doc.rust-lang.org/std/string/struct.String.html#method.len). Don't you love utf-8?

We're better off just calling `string.chars()` and succeeding iff there's exactly one char.

core/src/serde/de.rs
Stebalien: Maybe check validity (non-nan, etc.) here as well?

Stebalien: `(float as f32) as f64 == float` should be a sufficient "fits" test.

core/src/serde/de.rs
```diff
@@ -0,0 +1,693 @@
...
+    }
+
+    fn deserialize_struct<V: de::Visitor<'de>>(
+        self,
+        _name: &str,
+        _fields: &[&str],
+        visitor: V,
+    ) -> Result<V::Value, Self::Error> {
+        match self {
+            Self::Map(map) => visit_map(map, visitor),
```
Stebalien, line 449: Do we need to check the fields here?

core/src/serde/de.rs
```diff
@@ -0,0 +1,693 @@
...
+            )),
+        }
+    }
+
+    fn deserialize_tuple<V: de::Visitor<'de>>(
+        self,
+        _len: usize,
+        visitor: V,
+    ) -> Result<V::Value, Self::Error> {
+        self.deserialize_seq(visitor)
```
Stebalien: Do we need to check the length?

core/src/serde/de.rs
Stebalien: Yeah, I think we need a different deserializer. One that expects a call to (and only a call to) `deserialize_newtype_struct`.

core/src/serde/de.rs
Stebalien: See below, I think this is backwards. If we're told to visit a newtype struct, I think we need to turn around and try to deserialize into a CID. `BytesToCidVisitor` will deserialize any bytes into a CID.

core/src/serde/de.rs
```diff
@@ -0,0 +1,693 @@
...
+    }
+
+    // Heavily based on
+    // https://github.com/serde-rs/json/blob/95f67a09399d546d9ecadeb747a845a77ff309b2/src/value/de.rs#L446
+    fn deserialize_ignored_any<V: de::Visitor<'de>>(
+        self,
+        visitor: V,
+    ) -> Result<V::Value, Self::Error> {
+        drop(self);
+        visitor.visit_none()
```
Stebalien: You know, I wonder if this should be a `visit_unit`?

core/src/serde/de.rs
```diff
@@ -0,0 +1,693 @@
...
+
+fn visit_map<'de, V>(map: BTreeMap<String, Ipld>, visitor: V) -> Result<V::Value, SerdeError>
+where
+    V: de::Visitor<'de>,
+{
+    let mut deserializer = MapDeserializer::new(map);
+    visitor.visit_map(&mut deserializer)
+}
+
+fn visit_seq<'de, V>(list: Vec<Ipld>, visitor: V) -> Result<V::Value, SerdeError>
```
Stebalien, line 546: We're dropping the size hints here.

Stebalien (changes requested):
Starting with a deserialize review. I'll review serialize next.

core/src/serde/ser.rs
```diff
@@ -0,0 +1,486 @@
...
+        self,
+        _name: &'static str,
+        _variant_index: u32,
+        variant: &'static str,
+        value: &T,
+    ) -> Result<Self::Ok, Self::Error>
+    where
+        T: ser::Serialize,
+    {
+        let values = BTreeMap::from([(variant.to_owned(), value.serialize(self)?)]);
```
Stebalien, line 239: uber-nit: insert will be faster (this will get collected into a temporary array otherwise).

core/src/serde/ser.rs
```diff
@@ -0,0 +1,486 @@
...
+    fn serialize_field<T: ?Sized>(&mut self, value: &T) -> Result<(), Self::Error>
+    where
+        T: ser::Serialize,
+    {
+        self.vec.push(value.serialize(Serializer)?);
+        Ok(())
+    }
+
+    fn end(self) -> Result<Self::Ok, Self::Error> {
+        let map = BTreeMap::from([(self.name, Self::Ok::List(self.vec))]);
```
Stebalien, line 402: (insert would be faster)

core/src/serde/de.rs
```diff
@@ -0,0 +1,693 @@
...
+            #[inline]
+            fn visit_map<V>(self, mut visitor: V) -> Result<Self::Value, V::Error>
+            where
+                V: de::MapAccess<'de>,
+            {
+                let mut values = BTreeMap::new();
+
+                while let Some((key, value)) = visitor.next_entry()? {
+                    values.insert(key, value);
+                }
```
Stebalien, line 168: uber-nit: in cases like this, it might be faster to collect into a vector first, then construct the map from `vec.into_iter()`. Internally, the BTreeMap will:

1. Collect that iterator back into a `vec`, which is a no-op due to specialization.
2. Sort it.
3. Perform "bulk" insertion to construct the tree all at once.

Of course, that might not really be worth it. Up to you.

Stebalien (commented):
nits

core/src/serde/de.rs
vmx: I've put it into the documentation, that floats are converted with loss. I thought one might want to serialize into a `f32`, e.g. within a struct and don't really care about the loss of precision. But I can also see the point of only converting if it fits. I don't really have a strong opinion on that one, so I just made the change.

core/src/serde/de.rs
vmx: Good catch! It's embarrassing that I didn't catch that one myself ;)

core/src/serde/de.rs
vmx: I've to admit, I don't fully understand how `deserialized_ignored_any()` is actually triggered. I just don't understand its documentation. So I'm fine with `visit_unit()` if you think it's correct.

core/src/serde/de.rs
vmx: Do you mean, that there should be a `size_hint` implementation in the `de::SeqAccess` impl? This is what `serde_json` does.

core/src/serde/ser.rs
vmx: I hoped that it would be the reverse and that the compiler could optimize that one even better.

vmx (commented):
I've skipped the serializer nits for now. I guess I'll just leave them like that for now (I'm in the "optimize when things work and bottlenecks are identified camp :). Though it's always great to hear about such possible improvements.

One big thing I haven't addressed with the most recent changes is the `CidDeserializer` and the comments around whether things would work with a non-dag Serde codec. I'll look into those tomorrow.

vmx:
Oh, and most importantly I forgot to mention: thanks @Stebalien for your in-depth thoughtful review. It's amazing how much thought you put into this and how many things you spot that I've missed. I really also want to get this right. Thanks!

core/src/serde/de.rs
vmx: I've added a few more tests (the the most recent commit). `struct Wrapper(ByteBuf)`, does *not* deserialize a CID. If you still have concerns, please provide concrete tests (can be pseudo code), what I should test for.

vmx: > Hm. Well, we need to support deserializing IPLD -> IPLD...

~While I was looking into this, I was actually wondering about that. Do we really need that? Does a `deserialize_any` implementation make sense, or should we just remove it? Why would I want to deserialize `Ipld` into `Ipld`? I mean, I have an `Ipld` object already.~
`deserialized_any` is e.g. used when you'd want to deserialize `Ipld` into a `serde_json::Value`.

Stebalien: Generic code? Without that, we could end up with some strange bugs.

core/src/serde/de.rs
Stebalien: Personally, I'd rather force the user to specify that they're fine with a loss. They can always decorate the f32 with some custom deserializer that decodes into an f64 and casts.

core/src/serde/de.rs
Stebalien: Oh, I see. It literally doesn't matter. `deserialize_ignored_any` is just a hint to the deserializer that it shouldn't _actually_ bother decoding anything. The visitor will "go through the motions", but will ignore any results. E.g., if you ask it to visit a map, it'll do that then throw away the results.

So calling `visit_something_free` is the right way here (none, unit, etc.). Let's leave it as-is.

core/src/serde/de.rs
Stebalien: Oh, never mind. I noticed we were dropping `_len` in https://github.com/ipld/libipld/blob/ad768fe128a76c2c1ea0fcb40bbcc1ff23644ae8/core/src/serde/de.rs#L696, but we don't actually need that (the list tells us what we need).

core/src/serde/ser.rs
Stebalien: Allocations (side effects) can make optimizations a bit difficult, but I haven't tested.

core/src/serde/de.rs
```diff
@@ -0,0 +1,741 @@
...
+            )),
+        }
+    }
+
+    fn tuple_variant<V>(self, _len: usize, visitor: V) -> Result<V::Value, Self::Error>
+    where
+        V: de::Visitor<'de>,
+    {
+        match self.0 {
+            Some(Ipld::List(v)) => visit_seq(v, visitor),
```
Stebalien: Check that `_len == v.len()`?

core/src/serde/de.rs
```diff
@@ -0,0 +1,741 @@
...
+        }
+    }
+
+    fn deserialize_tuple_struct<V: de::Visitor<'de>>(
+        self,
+        _name: &str,
+        _len: usize,
+        visitor: V,
+    ) -> Result<V::Value, Self::Error> {
+        self.deserialize_seq(visitor)
```
Stebalien: call `self.deserialize_tuple` and pass the length.

core/src/serde/de.rs
```diff
@@ -0,0 +1,741 @@
...
+        self,
+        _name: &str,
+        fields: &[&str],
+        visitor: V,
+    ) -> Result<V::Value, Self::Error> {
+        match self {
+            Self::Map(map) => {
+                let keys: Vec<_> = map.keys().collect();
+                let mut fields_sorted = fields.to_vec();
+                fields_sorted.sort_unstable();
```
Stebalien: So... I think I might have lied here. I _think_ we need to just treat the fields as a hint. Otherwise, we can't accept objects with extra fields. The decision to allow unrecognized fields should lie with the visitor.

core/src/serde/de.rs
```diff
@@ -0,0 +1,741 @@
...
+            Self::Float(float) => {
+                if float.is_finite() {
+                    if (float as f32) as f64 == float {
+                        visitor.visit_f32(float as f32)
+                    } else {
+                        error("`Ipld::Float` cannot be deserialized to `f32`, without loss of precision`")
+                    }
+                } else {
+                    error(format!("`Ipld::Float` must be a finite number, not infinity or NaN, input was `{}`", float))
+                }
```
Stebalien: nit:

```suggestion
                if !float.is_finite() {
                    error(format!("`Ipld::Float` must be a finite number, not infinity or NaN, input was `{}`", float))
                } else if (float as f32) as f64 != float {
                    error("`Ipld::Float` cannot be deserialized to `f32`, without loss of precision`")
                } else {
                    visitor.visit_f32(float as f32)
                }
```

core/src/serde/de.rs
Stebalien: > I've added a few more tests (the the most recent commit). struct Wrapper(ByteBuf), does not deserialize a CID. If you still have concerns, please provide concrete tests (can be pseudo code), what I should test for.

Hm. Something still seems strange. I'll try to create a test.

core/src/serde/de.rs
```diff
@@ -0,0 +1,741 @@
...
+                            "Only `Ipld::Map`s with a single key can be deserialized to `enum`, input had no keys"
+                        );
+                    }
+                };
+                // Enums are encoded in IPLD as maps with a single key-value pair
+                if iter.next().is_some() {
+                    return error(
+                        "Only `Ipld::Map`s with a single key can be deserialized to `enum`, input had more keys"
+                    );
+                }
```
Stebalien, line 507: ot: what I wouldn't give to have a `try_collect`:

```rust
let [(variant, value)] = map.into_iter().try_collect()?;
```

core/src/serde/de.rs
Stebalien: Ah, it's because we're not calling "deserialize any".

core/src/serde/de.rs
Stebalien: This also panics when attempting to decode a CID into an untagged union.

core/src/serde/de.rs
Stebalien: Ok, so, to trigger this, some code would need to:

1. Call `deserialize_any`.
2. Get called by `visit_newtype_struct`.
3. Call `deserialize_bytes`.

At the moment, I can only get step 3 to call back into `deserialize_any` (which panics). I'm not sure if anyone is turning around and calling `deserialize_bytes` in this case, but there's no reason they _can't_.

However, it would still be safer to expect a call to `deserialize_newtype_struct` so we can _check_ that we're deserializing into the correct thing. I.e.:

1. This function passes a deserializer that _only_ succeeds if called with `deserialize_newtype_struct` with the special CID string.
2. `visit_newtype_struct` on the `Ipld` type calls back into `deserialize_newtype_struct`.

We won't get infinite recursion because we'd be passing a custom deserializer.

core/src/serde/de.rs
Stebalien: https://github.com/ipld/libipld/pull/141

core/src/serde/de.rs
vmx: I still feel like I'm missing something. Are we talking about a call like this:

```rust
Ipld::deserialize(something_that_impls_Deserializer);
```

You'd want to deserialize from some codec into `Ipld`. This would usually be some blob that implements a DAG codec. So now your question is: "could we get any strange results, when some random `Deserializer` is used?". I don't know why you would do that, but let's say you do.

Your `Deserializer` implementation would need to have a code path that calls `visit_newtype_struct`. Then that `Deserializer` also needs to have a `deserialize_bytes` implementation. That could of course call any `visit_*` it likes. But luckily we pass in the [`BytesToCidVisitor`](https://github.com/multiformats/rust-cid/blob/6cf0f6768e5b9e12f802e23bb63673d729898eef/src/serde.rs#L38-L68), which only implements `visit_bytes` and `visit_seq`. So all other `visit_*` calls would error, no harm here. When it calls `visit_bytes` or `visit_seq` we get some bytes or a sequence. If those form by coincidence a valid CID, we would return that CID. Else we would again error.

To conclude: a random `Deserializer` implementation is likely to error. Only if it leads to a byte sequence that leads to a valid CID, a CID is returned. And I would say if it does, it was likely a codec that was crafted to support CIDs.

core/src/serde/de.rs
vmx: I've changed it to `visit_unit()`, but I can also revert it.

Stebalien: I'd leave it as is.

core/src/serde/de.rs
Stebalien: So "random bytes" aren't the issue, just assume that the attacker can put bytes forming a CID anywhere. The constraint is that some `links` function that extracts all CIDs in some object should list the exact same CIDs we get when we decode that object.

Basically:

- Assume the programmer is well-meaning but may not have the same mental model as you.
- Assume that input data is malicious.

core/src/serde/de.rs
vmx: Cool, then this can be removed as Serde takes care of erroring in the other case like not matching or missing fields.

core/src/serde/de.rs
vmx: Resolved, but later comment. We don't need special checks here, missing or wrong fields will error, tests are there to cover that case.

core/src/serde/de.rs
vmx: > 2. `visit_newtype_struct` on the `Ipld` type calls back into `deserialize_newtype_struct`.

This is not clear to me. To which `deserialize_newtype_struct` is it calling back, to the original one?

> We won't get infinite recursion because we'd be passing a custom deserializer.

I lost count, how many deserializers would we have? Changing the `CidDeserializer` to one that accepts `deserialize_newtype_struct` only, or is this yet another one.

When I was developing this, I always hit the problem that eventually I need to call `deserialize_bytes` as I haven't found any other way to pass on the bytes of the CID into a visitor, which then needs to construct the `Cid`. Though it can well be that there is a way.

core/src/serde/de.rs
Stebalien: I'm setting the bar to "we can prove to ourselves that this is correct" not "we can't think of a way for this to go wrong".

core/src/serde/de.rs
Stebalien: > To which deserialize_newtype_struct is it calling back, to the original one?

A new one, unfortunately.

---

But I don't think that works. `visit_newtype_struct` really shouldn't call `deserialize_newtype_struct`. It should always be the other way around. The problem we're running into here is that we're calling `visit_newtype_struct` from `deserialize_any` where we don't have any type information.

Stebalien (approved):

PR #141, opened 2022-02-24
test: demonstrate panic when decoding into an untagged union

Stebalien:


vmx:
I've pushed a fix together with this test case to https://github.com/ipld/libipld/pull/140. Thanks for the test case!

PR #143, opened 2022-02-25
WIP: represent CID as enum and not as newtype struct

vmx:
Changing from newtype struct to enum should hopefully make things
more robust.

### DIFF
--- a/core/src/serde/de.rs
+++ b/core/src/serde/de.rs
@@ -4,7 +4,7 @@ use core::{convert::TryFrom, fmt};
 use cid::serde::{BytesToCidVisitor, CID_SERDE_PRIVATE_IDENTIFIER};
 use cid::Cid;
 use serde::{
-    de::{self, IntoDeserializer},
+    de::{self, IntoDeserializer, VariantAccess},
     forward_to_deserialize_any,
 };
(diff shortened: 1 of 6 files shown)


vmx:
The corresponding change for cbor: https://github.com/vmx/cbor/pull/5

core/src/serde/de.rs
```diff
@@ -464,31 +547,30 @@ impl<'de> de::Deserializer<'de> for Ipld {
...
     fn deserialize_enum<V: de::Visitor<'de>>(
         self,
-        _name: &str,
-        _variants: &[&str],
+        name: &str,
+        variants: &[&str],
         visitor: V,
     ) -> Result<V::Value, Self::Error> {
-        let (variant, value) = match self {
+        if name == CID_SERDE_PRIVATE_IDENTIFIER && variants == [CID_SERDE_PRIVATE_IDENTIFIER] {
```
Stebalien, line 564: possible extension: allow the variants to be arbitrary IPLD kinds.

Stebalien (approved):
To me, this seems like it's strictly less likely to get misinterpreted. But the final decision is up to you.

PR #144, opened 2022-03-07
fix: support deserializing unit

vmx:


vmx:
@Stebalien I can't recall why we wanted this to be an error.

Also: would we want the [`IpldVisitor`](https://github.com/ipld/libipld/blob/0395e9bae8aabec3f43f8fed01151a973de2502c/core/src/serde/de.rs#L66) to be able to visit `unit` (currently it doesn't)?

Stebalien (approved):

Stebalien:
Actually, I may have pressed that button too fast. Don't we want to only deserialize into a unit if self is null?

Stebalien:
We originally decided _not_ do do anything because it was unclear what the IPLD serialization of "unit" should be. I preferred to make it explicit.

Stebalien:
Fix in https://github.com/ipld/libipld/pull/145

PR #145, opened 2022-03-08
fix: only deserialize unit from null

Stebalien:


core/tests/serde_deserialize.rs
```diff
@@ -19,7 +19,7 @@ fn ipld_deserialize_null() {
 #[test]
 #[should_panic(expected = "invalid type")]
 fn ipld_deserialize_null_not_as_unit() {
-    let ipld = Ipld::Null;
+    let ipld = Ipld::Bool(true);
     assert_de_tokens(&ipld, &[Token::Unit]);
 }
```
vmx: This whole test should just be removed.

Stebalien: You sure? We're asserting that this _doesn't_ work.

Stebalien: Oh, I see.

vmx (approved):

PR #149, opened 2022-08-19
Expose serializer

appcypher:
This PR exposes the ipld `Serializer` struct because there are times when one needs to create a custom serialization pipeline. In our case, we needed the serialization to be async.

https://github.com/wnfs-wg/rs-wnfs/blob/8471980856716bbdf54b77cc01fe4b7c350e559b/crates/fs/traits.rs#L67-L73

### DIFF
--- a/core/src/serde/mod.rs
+++ b/core/src/serde/mod.rs
@@ -6,7 +6,7 @@ mod de;
 mod ser;
 
 pub use de::from_ipld;
-pub use ser::to_ipld;
+pub use ser::{to_ipld, Serializer};
 
 #[cfg(test)]
 mod tests {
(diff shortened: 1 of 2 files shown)


vmx (approved):

matheus23:
@vmx what is needed to get this merged & released? Can we help?
Some checks fail, but I suspect that's not due to the changes in this PR, right?

vmx:
@matheus23 that second I was looking into the CI failures. I agree it seems unrelated, but fixing those seem to be more work than I thought. Hence I've opened https://github.com/ipld/libipld/issues/150.

So I'll merge this one, but before a release I'd like the CI to be green.

vmx:
I should've been more clear:

> @vmx what is needed to get this merged & released? Can we help?

I don't have the time to look into that at the moment, so if anyone could help with #150, that would be great!

Issue #150, opened 2022-08-22
Get CI back to green


vmx:
Currently the `master` branch on CI is not green. Fix all outstanding issues.

matheus23:
I think this can be closed :)

@vmx Can you cut a release of `libipld`? This is currently blocking us for releasing a new version of `rs-wnfs` :eyes:

vmx:
I had to talk with others involved about the release. It's done now.

PR #151, opened 2022-08-22
Fix lint-rust action

matheus23:
Work on #150 

This fixes some issues introduced by compiling with rust 1.63 instead of 1.62.

I can't quite figure out the `no_std` problems though. It seems like there's some dependency that re-enables the `std` feature on `serde`, but I don't know which and how to debug it. Here's the output I get from building for a no_std target:

```
$ cargo build --no-default-features --target thumbv6m-none-eabi --manifest-path core/Cargo.toml
   Compiling data-encoding v2.3.2
   Compiling serde v1.0.144
   Compiling anyhow v1.0.62
error[E0463]: can't find crate for `std`
  |
  = note: the `thumbv6m-none-eabi` target may not support the standard library
  = note: `std` is required by `serde` because it does not declare `#![no_std]`
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "libipld"
 version = "0.13.1"
 authors = ["David Craven <david@craven.ch>"]
-edition = "2018"
+edition = "2021"
 license = "MIT OR Apache-2.0"
 description = "library for dealing with ipld"
 repository = "https://github.com/ipld/libipld"
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -2,7 +2,7 @@
 name = "libipld-core"
 version = "0.13.1"
 authors = ["David Craven <david@craven.ch>"]
-edition = "2018"
+edition = "2021"
 license = "MIT OR Apache-2.0"
 description = "Base traits and definitions used by ipld codecs."
 repository = "https://github.com/ipfs-rust/rust-ipld"
--- a/core/tests/serde_deserializer.rs
+++ b/core/tests/serde_deserializer.rs
@@ -56,6 +56,7 @@ where
 
 #[test]
 #[allow(clippy::unit_cmp)]
+#[allow(clippy::let_unit_value)]
 fn ipld_deserializer_unit() {
     let unit = ();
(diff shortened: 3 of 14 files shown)


vmx:
> but I don't know which and how to debug it.

`cargo tree --edges features` might help.

matheus23:
This *seems* to be caused by the `serde_test` `devDependency`: It seems like `std` can't be turned off for `serde_test`.
According to `cargo tree --no-default-features -i serde --edges features`, `serde_bytes` and `serde_json` also cause this, but I can turn off default features for them.

I'm really no expert when it comes to rust's build systems. Any idea on how to resolve this? It *seems* similar to this issue I found: https://github.com/serde-rs/serde/issues/1535

vmx:
It also fails on CI, but I can't reproduce it locally :-/. What I would've tried would've been upgrading the crate to Rust edition 2021 (https://doc.rust-lang.org/edition-guide/editions/transitioning-an-existing-project-to-a-new-edition.html#transitioning-an-existing-project-to-a-new-edition).

matheus23:
I think when I ran this locally I accidentally changed some `build.rs` generated files which of course aren't checked in.
They were generated by `prost-build`, so I upgraded that an it *should* now compile successfully.

vmx:
`prost` v0.11 [doesn't bundle `protoc` anymore](https://github.com/tokio-rs/prost#protoc). This means that if we upgrade, the CI would need to be adapted. I'd be OK with not upgrading.

matheus23:
Alright. I
- upgraded the edition and fixed resulting errors in the proc macros,
- downgraded prost-build to 0.10 (and upgraded prost to 0.10 just to be sure)
- added an `#![allow(clippy...)]`, otherwise clippy would complain about generated code (:man_facepalming:)

The CI looks green on my fork, I'm pretty confident it'll be green here too @vmx

matheus23:
:green_circle: :tada: 
Would love for this to get released soon :)

vmx (approved):

vmx:
Thanks a lot @matheus23 for working through all this and getting CI back to green!

PR #153, opened 2022-09-08
Remove Store trait, test implementations, and some dependencies

rklaehn:
First of all: I really like the way the Store trait works. It is something we came up with when working at Actyx for ipfs-embed.

But it seems to me that the Store trait and its implementations don't really belong in libipld. libipld should be purely about the ipld data format, not about how to write an ipfs for it.

So this removes it. It does seem to be pretty isolated. I am not aware of anybody other than ipfs-embed using this kind of approach for a store. But I don't know. Maybe it would be safer to first deprecate it for some versions.


vmx (commented):
I think storage in general shouldn't be tightly integrated with this library. It could live outside as different projects have different needs. I agree that if the only use is ipfs-embed, it should rather be moved there.

Does anyone else have objections removing the code?

dignifiedquire:
👍 independent of agreeing with the interface or not, it should be separate from libipld

rklaehn:
As far as I understaind, the version of ipfs-embed that is currently in production can not update to the latest libipld anyway for different reasons. Is that correct, @rkuhn ? In that case I think it would be OK to merge this.

rkuhn:
Yes, I also find it slightly weird to pull in the store trait from this library.

rklaehn:
@vmx so, good to go?

vmx (approved):
Yes!

PR #159, opened 2022-11-04
add dagcbor::validate() function.

### LINKED ISSUE #158: dag-cbor: add ability to validate input without fully deserializing
To implement https://github.com/filecoin-project/FIPs/pull/483 securely in the FVM, we need to validate that entry values are well-formed DAG-CBOR payloads. All we need is to perform syntatical validation without incurring in any serde costs/overheads.

raulk:
This utility function validates that a supplied input is syntatically valid DAG-CBOR, while attempting to deserialize as little as possible.

- It has to deserialize numbers to validate they're minimally encoded.
- It validates that map keys are strings, and map values are valid DAG-CBOR.
- It validates that list elements are valid DAG-CBOR.
- It does not validate UTF-8 encoding of strings.

Closes https://github.com/ipld/libipld/issues/158.

We may want to remove the recursion here to improve security against untrusted inputs.

### DIFF
--- a/dag-cbor/src/lib.rs
+++ b/dag-cbor/src/lib.rs
@@ -10,6 +10,7 @@ pub mod cbor;
 pub mod decode;
 pub mod encode;
 pub mod error;
+pub mod validate;
 
 /// CBOR codec.
 #[derive(Clone, Copy, Debug, Default, PartialEq, Eq, PartialOrd, Ord)]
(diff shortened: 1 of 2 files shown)


vmx:
I haven't really looked at the code yet, just a general note: for FVM we don't use the DAG-CBOR implementation from this library, but we use https://github.com/ipld/serde_ipld_dagcbor. So if validation functionality would be added, it should be added there instead.

PR #161, opened 2022-11-09
Get rid of prost-build by vendoring the DAG protobuf representation

fabricedesre:
@vmx as we discussed this removed the need to generate the Rust representation from the .proto file since it won't change. I left it in for reference.

### DIFF
--- a/core/src/serde/mod.rs
+++ b/core/src/serde/mod.rs
@@ -29,7 +29,7 @@ mod tests {
     where
         T: Serialize + DeserializeOwned + PartialEq + fmt::Debug,
     {
-        let encoded: Ipld = to_ipld(&data).unwrap();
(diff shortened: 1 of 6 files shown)


fabricedesre:
Thanks! Can you do a release on crates.io ?

vmx:
This is a breaking change release due to https://github.com/ipld/libipld/commit/8ff8bd6021047bb5686ee72d08944cd56a23ba64. A breaking rust-cid release (due to upgrading quickcheck` is upcoming) and I'd like to add that version upgrade (together with the rust-multihash one) in the breaking change release. I'm sorry that this takes so long to land in a release.

vmx:
@fabricedesre sorry for taking so long, there's now a new release out.

PR #165, opened 2022-11-16
Make the deserializer human-unreadable

samuelburnham:
Currently, the Serializer sets `is_human_readable` to false, but the Deserializer implicitly sets it to true. This causes issues when other types try to use serde based on this value, as in https://github.com/zcash/pasta_curves/pull/48, so this PR makes them consistently false.


vmx (approved):
Good catch, thanks!

PR #166, opened 2022-11-17
Update `cid` and `multihash` in `core`

tyshko5:


tyshko5:
@Stebalien @vmx Could you give an approval so it could pass to the checks?

vmx (changes requested):
Please update `cid` and `multihash` for all crates of this project.

Stebalien (approved):

tyshko5:
@Stebalien @vmx Are you going to release a new version for `libipld` and it's subcrates? That would be very helpful for updating a lot of crates that are depending on it.

vmx:
Yes, I was waiting for this PR (also other folks really want a release). Though I wondered if we can also get the quickcheck one in before doing a release.

vmx:
After a second thought, based on the PR comments on the quickcheck change, it might still need some work and it also is not a breaking change. Hence releasing a new libpld version now.

vmx:
New 0.15 release is out.

PR #172, opened 2022-12-12
Adds Eq instance for Ipld

### LINKED ISSUE #171: Question: Why isn't `Eq` implemented for `Ipld` data type?
I see that `PartialEq` is implemented (via derivation), however, I don't see any annotation in the source for why `Eq` cannot be implemented. Does this have something to do with `Null`?

ProofOfKeags:
Fixes #171

### DIFF
--- a/core/src/ipld.rs
+++ b/core/src/ipld.rs
@@ -36,6 +36,8 @@ pub enum Ipld {
     Link(Cid),
 }
 
+impl Eq for Ipld {}
+
 impl fmt::Debug for Ipld {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         if f.alternate() {
--- a/core/src/ipld.rs
+++ b/core/src/ipld.rs
@@ -9,8 +9,41 @@ use alloc::{
 };
 use core::fmt;
 
-use crate::cid::Cid;
 use crate::error::TypeError;
+use crate::{cid::Cid, error::TypeErrorType};
+
+#[derive(Debug, Clone, PartialEq)]
+/// Floating point type that excludes NaN, Infinity, -Infinity, and subnormal values
+pub struct FiniteFloat(f64);
+impl FiniteFloat {
+    /// Constructs a FiniteFloat from an f64 if it is not NaN, Infinite, or subnormal
+    pub fn from_normal_float(normal: f64) -> Option<Self> {
+        if normal.is_normal() {
+            Some(FiniteFloat(normal))
+        } else {
+            None
+        }
+    }
+
+    /// Converts to f64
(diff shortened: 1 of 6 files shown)


core/src/ipld.rs
```diff
@@ -9,8 +9,41 @@ use alloc::{
 };
 use core::fmt;
 
-use crate::cid::Cid;
 use crate::error::TypeError;
+use crate::{cid::Cid, error::TypeErrorType};
+
+#[derive(Debug, Clone, PartialEq)]
+/// Floating point type that excludes NaN, Infinity, -Infinity, and subnormal values
```
rkuhn, line 16: Out of curiosity: why forbid infinite and subnormal values? They could easily participate in an `impl Ord + Eq`.

vmx: They are excluded by the IPLD data model if you use IEEE floats: https://github.com/ipld/ipld/blob/master/docs/data-model/kinds.md#float-kind

vmx (commented):
Wrapping it in a newstruct has a high impact on downstream users. While I think it might be a way to do this correctly, it needs more thought and input from other downstream users. So please don't expect this change to be merged quickly (or at all).

core/src/ipld.rs
rkuhn: Yes, but there is no further explanation why these particular values are outlawed, hence my curiosity.

vmx: The IPLD Data Model is intentionally kept abstract. E.g. we have "integer" and "float" kinds. Within the IPLD Data Model, it's not specified how they are represented. Integers could e.g. be arbitrary precision ones in some implementations, 64 or 128-bit integers or even IEEE doubles (hello JS). The same is true for floats. I would expect that they are usually implemented as IEEE doubles. But you could also think of implementations that uses fractions instead. As we want to keep that option open, we non numeric things. There is also more information on floats at https://github.com/ipld/ipld/blob/015ff80aea78f588b981c6cef7f1e9bd59d31f48/design/tricky-choices/numeric-domain.md.

ProofOfKeags:
Would you prefer a different implementation? I don't necessarily think it needs to be newtyped but that's the way we can make explicit what properties are checked. The issue is that Rust has no way of hiding data constructors (in this case Float), so the only way to ensure that invalid IPLD's cant be directly constructed is to bury it in a newtype. Personally I don't think this is needed, since clearly this is already an issue as it stands and that a failure to adhere to the requirements of the data model can just mean that Eq blows up. I do think it is important that Eq is implemented for IPLD though. Since deterministic serialization implies a total equality function, not having Eq implemented is a real head scratcher at minimum and a real thorn at maximum.

PR #174, opened 2023-01-11
fix: improve codec spec compliance

vmx:
The DAG-CBOR, DAG-JSON and DAG-PB codecs are not more spec compliant. New negative test fixtures were added to https://github.com/ipld/codec-fixtures, with this change all pass.

BREAKING CHANGE: DAG-CBOR, DAG-JSON and DAG-PB encoding and decoding is stricter.

In DAG-CBOR and DAG-JSON now error when duplicated map keys with the same map are encountered.

DAG-PB is now stricter on encoding and decoding. For example are links and data not allowed to be interleaved, they have to be one after another (in any order).

When encoding `Ipld` to DAG-PB, the types now need to match, links need to be ordered.

The DAG-PB will error if unknown data is encountered during decoding.

---

I've made it maximally strict. If there are problems with that let me know. I personally prefer making things strict first and relaxing them if it makes sense later on (so we don't need many breaking changes).

The corresponding codec fixtures PR is at https://github.com/ipld/codec-fixtures/pull/85.

### DIFF
--- a/dag-cbor/src/error.rs
+++ b/dag-cbor/src/error.rs
@@ -115,3 +115,8 @@ pub struct UnexpectedEof;
 #[derive(Debug, Error)]
 #[error("Invalid Cid prefix: {0}")]
 pub struct InvalidCidPrefix(pub u8);
+
+/// A duplicate key within a map.
+#[derive(Debug, Error)]
+#[error("Duplicate map key.")]
+pub struct DuplicateKey;
--- a/core/src/serde/de.rs
+++ b/core/src/serde/de.rs
@@ -164,7 +164,10 @@ impl<'de> de::Deserialize<'de> for Ipld {
                 let mut values = BTreeMap::new();
 
                 while let Some((key, value)) = visitor.next_entry()? {
-                    values.insert(key, value);
+                    let prev_value = values.insert(key, value);
+                    if prev_value.is_some() {
+                        return Err(de::Error::custom("Duplicate map key"));
+                    }
                 }
 
                 Ok(Ipld::Map(values))
--- a/dag-cbor/src/decode.rs
+++ b/dag-cbor/src/decode.rs
@@ -1,8 +1,8 @@
 //! CBOR decoder
 use crate::cbor::{Major, MajorKind, F32, F64, FALSE, NULL, TRUE};
 use crate::error::{
-    InvalidCidPrefix, LengthOutOfRange, NumberNotMinimal, NumberOutOfRange, UnexpectedCode,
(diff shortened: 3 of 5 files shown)


dag-cbor/src/error.rs
```diff
@@ -115,3 +115,8 @@ pub struct UnexpectedEof;
 #[derive(Debug, Error)]
 #[error("Invalid Cid prefix: {0}")]
 pub struct InvalidCidPrefix(pub u8);
+
+/// A duplicate key within a map.
+#[derive(Debug, Error)]
+#[error("Duplicate map key.")]
+pub struct DuplicateKey;
```
dignifiedquire, line 122: would be nice to inform which key is duplicate

vmx: Yes, it would be nice, but it's a pain to implement, hence I didn't do it.

rvagg:
On the enforcement of sorted-Links on encode thing, that's a call that should be dependent on the model being used in Rust.

* It's enforced in JS so we get that property that an object that does a round-trip through a codec is the same and you don't end up with an array that's all jumbled up (and a deepEqual would work). But we also have a `prepare()` call there to take very sloppy object forms and munge them into that strict form to help out with the pain of strictness.
* It's not enforced in Go: https://github.com/ipld/codec-fixtures/blob/d66748515e65cefee3c57fc742363ee90e6a1f0f/go/special_cases.go#L10-L11 because it doesn't make quite as much sense there thanks to the `Node` data model forms and the opinions embedded in that way of representing things.

And you could argue either way with both of those calls; and perhaps that case shouldn't have been included in the negative fixtures at all. So I wouldn't consider this a super important rule if it doesn't make sense here like it does in JS.

_(EDIT: to clarify, the resulting encoded form should still be sorted, that ends up being kind of important for various traversal cases, I'm talking about enforcing what you receive to be encoded)_

dag-cbor/src/error.rs
vmx: I should have written a more elaborate reply: The problem is that it's generic over `K`. So I'd need to add the `Display` trait. That would then bleed into other things. I found it a nice to have, but not important enough to figure out all the details.

vmx:
Thanks @rvagg for the details here. I'm OK with having the ordering in place, but I don't feel strongly about it. So if anyone has concerns/objections, I'm happy to remove it (it's still enforced on the encoding step). Generally I prefer "error early". But if it turns out to be a pain point, we can also make it less strict later.

dignifiedquire (approved):

PR #176, opened 2023-01-23
chore: remove consolidate-pushes from cargo-release config

vmx:
Since cargo-release v0.22 the `consolidate-pushes` option is always enabled, hence it was removed. Update the Cargo.toml so that we can do releases with cargo-release newer than v0.22.


vmx:
I dare to merge this one without review as it's really only about details on how to do the release.

PR #177, opened 2023-01-27
feat: add DAG-JOSE implementation

nathanielc:
This is an implementation of DAG-JOSE. There are also a few minor changes to existing code.

* An update to the dag-cbor-derive code in order to allow usage of the macro from within this repo.
* Add a `Bytes` type to dag-cbor to make declaring bytes within a struct more straightforward.
* A dag-jose feature

There are some things left to do before I would consider this PR ready. However I would like feedback on the general design first. You can see its usage in the fixtures.rs test case.

* Provide complete examples in docs
* Determine what to do about the fixture tests. We likely want to move that logic to https://github.com/ipld/codec-fixtures
* Add Ipld conversion to the types to avoid round-tripping through CBOR for JSON encoding.

### DIFF
--- a/dag-cbor/src/lib.rs
+++ b/dag-cbor/src/lib.rs
@@ -36,6 +36,11 @@ pub trait DagCbor: Encode<DagCborCodec> + Decode<DagCborCodec> {}
 
 impl<T: Encode<DagCborCodec> + Decode<DagCborCodec>> DagCbor for T {}
 
+/// Bytes is sequence of byte values.
+///
+/// Implements Encode and Decode to/from CBOR byte strings
+pub type Bytes = Box<[u8]>;
+
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/dag-jose/src/lib.rs
+++ b/dag-jose/src/lib.rs
@@ -1,0 +1,400 @@
+//! Jose codec.
+//! TODO
+#![deny(missing_docs)]
+#![deny(warnings)]
+
+mod codec;
+mod error;
+
+use std::collections::BTreeMap;
+
+use libipld_cbor::DagCborCodec;
+use libipld_core::cid::Cid;
+use libipld_core::codec::Codec;
+use libipld_core::codec::{Decode, Encode};
+use libipld_core::error::UnsupportedCodec;
+use libipld_core::ipld::Ipld;
+use libipld_json::DagJsonCodec;
+
+use codec::{Decoded, Encoded};
+
+use crate::{
+    codec::{DecodedRecipient, DecodedSignature, EncodedRecipient, EncodedSignature},
+    error::Error,
+};
+
+/// DAG-JOSE codec
+#[derive(Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
+pub struct DagJoseCodec;
+
+impl Codec for DagJoseCodec {}
+
+impl From<DagJoseCodec> for u64 {
+    fn from(_: DagJoseCodec) -> Self {
+        // Multicode comes from here https://github.com/multiformats/multicodec/blob/master/table.csv
+        0x85
+    }
+}
+
+impl TryFrom<u64> for DagJoseCodec {
+    type Error = UnsupportedCodec;
+
+    fn try_from(_: u64) -> core::result::Result<Self, Self::Error> {
+        Ok(Self)
+    }
+}
+
+impl Encode<DagJoseCodec> for Ipld {
+    fn encode<W: std::io::Write>(&self, _c: DagJoseCodec, w: &mut W) -> anyhow::Result<()> {
+        self.encode(DagCborCodec, w)
+    }
+}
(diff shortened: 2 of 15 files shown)


dag-cbor/src/lib.rs
```diff
@@ -36,6 +36,11 @@ pub trait DagCbor: Encode<DagCborCodec> + Decode<DagCborCodec> {}
 
 impl<T: Encode<DagCborCodec> + Decode<DagCborCodec>> DagCbor for T {}
 
+/// Bytes is sequence of byte values.
+///
+/// Implements Encode and Decode to/from CBOR byte strings
+pub type Bytes = Box<[u8]>;
```
dbcfd, line 42: I'm a little confused as to why this is needed. It seems like Vec<u8> is covered. Also why Box<[u8]> and not Vec<u8>.

dag-jose/src/lib.rs
```diff
@@ -0,0 +1,400 @@
...
+    use libipld_core::{cid::Cid, codec::assert_roundtrip};
+    use libipld_macro::ipld;
+
+    fn fixture_jws() -> (Box<[u8]>, Box<[u8]>, Box<[u8]>) {
+        let payload =
+            base64_url::decode("AXESIIlVZVHDkmZ5zFLHLhgqVhkFakcnQJ7pOibQWtcnyhH0").unwrap();
+        let protected = base64_url::decode("eyJhbGciOiJFZERTQSJ9").unwrap();
+        let signature =  base64_url::decode("-_9J5OZcl5lVuRlgI1NJEzc0FqEb6_2yVskUaQPducRQ4oe-N5ynCl57wDm4SPtm1L1bltrphpQeBOeWjVW1BQ").unwrap();
+        (
+            payload.into_boxed_slice(),
```
dbcfd, line 302: Seems like this should just be to_vec().

dag-jose/src/lib.rs
```diff
@@ -0,0 +1,400 @@
...
+            protected.into_boxed_slice(),
+            signature.into_boxed_slice(),
+        )
+    }
+    fn fixture_jws_base64(
+        payload: &Box<[u8]>,
```
dbcfd, line 308: And this should just take `&[u8]`

dag-cbor/src/lib.rs
nathanielc: It stems from here https://github.com/ipld/libipld/blob/master/dag-cbor/src/decode.rs#L297 and here https://github.com/ipld/libipld/blob/master/dag-cbor/src/decode.rs#L321

Using a `Vec<u8>` creates an IPLD list of u8 values, where using `Box<[u8]>` creates an IPLD byte sequence. 

We could change the cbor Decode implementations but this was the smallest change I could make to existing code. Additionally we could try and specialize the implementation but that required Rust nightly.

dbcfd: LIttle odd that they didn't handle Vec<u8> as a byte string, but is there a reason we need a ByteString here, and an array of u8 isn't sufficient. A large number of libraries work on Vec<u8> or bytes::Bytes, so to use this, they'll have to be regularly calling into_boxed_slice() to work with the types specified here. 

dag-pb actually uses bytes::Bytes. Oddly, it also returns a Box<[u8]> for the payload here https://github.com/ipld/libipld/blob/master/dag-pb/src/codec.rs#L78 rather than the Vec or a Bytes structure.

vmx:
This is not an in-depth review, but a few general notes.

The way I propose to use DAG-CBOR is via https://crates.io/crates/serde_ipld_dagcbor. It is based on Serde, this way you get access to the Serde ecosystem. Whenever you define Serde (de)serialization for some type, you can transform it into IPLD, that's a huge win. Given that, I hope we can [move away from the current `dag-cbor` crate in this repo and also from `dag-cbor-derive`](https://github.com/ipld/libipld/issues/164). The idea would be that instead of `dag-cbor-derive`, you would use Serde directly instead. I guess that most of its functionality is already be possible with Serde, the missing pieces could be implemented. The benefit of this approach is, that it would then work for other (future) `serde_ipld_dag*` codecs automatically and you won't need to re-implement things for a specific codec.

Given all that I propose keeping the DAG-JOSE codec in a separate repo. If changes to `dag-cbor-derive` are still needed, please open a separate PR for that. In regards to `dag-cbor` changes, generally I'd like to keep changes to a minimum as I consider it legacy, but I'm also not fully opposed changes if it enable users.

vmx:
> * Determine what to do about the fixture tests. We likely want to move that logic to https://github.com/ipld/codec-fixtures

It would be great to have DAG-JOSE in the `codec-fixtures` repo. There's also already some work there: https://github.com/ipld/codec-fixtures/pull/11.

nathanielc:
Thanks for the comments. I have implemented the changes against a new repo. https://github.com/ceramicnetwork/rust-dag-jose/pull/1

Issue #173, opened 2023-01-10
Implement TryFrom for builtin types


vmx:
I was facing the problem of decoding something into an `Ipld` enum, that contained a `Map`, whose fields I wanted to access. I knew the type of the field, in my case a string. So I wanted to do something like:

```rust
// `fixture` is an `Ipld::Map`
let name = String::try_from(fixture.get("name").unwrap()).unwrap();
```

But it's not implemented. It would be great if it would for all the Rust builtin types that map to the IPLD Data Model.

The workaround is doing something like:

```rust
if let Ok(Ipld::String(name)) = fixture.get("name") {
…
}
```

SionoiS:
Something like this?

```
macro_rules! derive_from_ipld {
    ($enum:ident, $ty:ty) => {
        impl TryFrom<Ipld> for $ty {
            type Error = TypeError;

            fn try_from(ipld: Ipld) -> Result<Self, Self::Error> {
                match ipld {
                    Ipld::$enum(value) => Ok(value),
                    _ => {
                        return Err(TypeError {
                            expected: TypeErrorType::$enum,
                            found: ipld.into(),
                        })
                    }
                }
            }
        }
    };
}
```
#179

PR #179, opened 2023-04-18
TryFrom for builtin types

SionoiS:
#173

### DIFF
--- a/core/src/convert.rs
+++ b/core/src/convert.rs
@@ -1,6 +1,7 @@
 //! Conversion to and from ipld.
-use crate::cid::Cid;
+use crate::error::TypeErrorType;
 use crate::ipld::Ipld;
+use crate::{cid::Cid, error::TypeError};
 use alloc::{
     borrow::ToOwned,
     boxed::Box,
@@ -8,6 +9,42 @@ use alloc::{
     string::{String, ToString},
     vec::Vec,
 };
+
+impl TryFrom<Ipld> for () {
+    type Error = TypeError;
+
+    fn try_from(ipld: Ipld) -> Result<Self, Self::Error> {
+        match ipld {
+            Ipld::Null => Ok(()),
+            _ => {
+                return Err(TypeError {
+                    expected: TypeErrorType::Null,
+                    found: ipld.into(),
+                })
+            }
+        }
+    }
+}
+
+macro_rules! derive_from_ipld {
+    ($enum:ident, $ty:ty) => {
+        impl TryFrom<Ipld> for $ty {
+            type Error = TypeError;
+
+            fn try_from(ipld: Ipld) -> Result<Self, Self::Error> {
+                match ipld {
+                    Ipld::$enum(value) => Ok(value),
+                    _ => {
+                        return Err(TypeError {
+                            expected: TypeErrorType::$enum,
+                            found: ipld.into(),
+                        })
+                    }
+                }
+            }
+        }
+    };
+}
 
 macro_rules! derive_to_ipld_prim {
     ($enum:ident, $ty:ty, $fn:ident) => {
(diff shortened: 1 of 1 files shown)


SionoiS:
I added a way to ```try_into()``` options.

I'm also thinking about ```AsRef<Ipld>``` IDK if it make sense yet.

vmx (commented):
Oh that looks really nice, thanks!

Would you mind adding some tests. I know we also previously haven't had tests, but at least some would be nice. Perhaps the original use case from #173? Though this will be merged even without tests.

vmx:
There's a Clippy warning unrelated to your change. I'll take care of it in a separate PR.

vmx:
PR for the Clippy warnings: https://github.com/ipld/libipld/pull/180 are review would be appreciated :)

SionoiS:
I added some tests and some doc on how casting is handled.

Your example in #173 would not work as is because ```get()``` return a reference.

core/src/convert.rs
```diff
@@ -1,6 +1,11 @@
 //! Conversion to and from ipld.
-use crate::cid::Cid;
+//!
+//! Please note conversion use the ```as``` keyword under the hood.
```
vmx: That's a good note. I haven't put much thought into it before. Would it be possible to just use `try_from()` for those conversions instead?

SionoiS: Error handling would be tricky as the error type for ```ipld.try_into()``` and ```TryFrom<Ipld>``` would have to be convertible into each other. Possible to do surely but not without changes to the error handling. Maybe create a new issue?

vmx: Hmm. Thinking from the users perspective. I probably wouldn't read the module level documentation and would expect the `TryFrom` doing what I'd expect it to do: converting things losslessly if there is no error. I feat that using `as` would lead to hard to find bugs for people using the library. E.g. "why is this integer value 524 suddenly 12". That would happen if it was converted to an u8.

SionoiS: True!

I'm used to ```this_error``` for error handling IDK how to do it with anyhow. I'll look into it.

SionoiS:
What to do with f32 there's no way to go from f64 -> f32 except with ```as```?

vmx:
> What to do with f32 there's no way to go from f64 -> f32 except with `as`?

I suggest not implementing it for `f32`. If someone needs it, they would need to convert first to `f64` and then be on their on doing it whichever way they like.

core/src/convert.rs
```diff
@@ -9,6 +13,60 @@ use alloc::{
...
+            _ => Err(TypeError {
+                expected: TypeErrorType::Null,
+                found: ipld.into(),
+            }
+            .into()),
+        }
+    }
+}
+
+macro_rules! derive_from_ipld_option {
```
vmx: I should've spotted that earlier. What about renaming both to `…try_from…` instead of just `try`? I don't have strong opinions about it, if you prefer keeping it like that, it's OK for me.

SionoiS: Right! try_from make more sense.

SionoiS:
I just noticed ```no-std``` doesn't build.

Should I gate the ```try_from``` with std? Seams required for the error conversion.

core/src/convert.rs
```diff
@@ -29,36 +87,212 @@ macro_rules! derive_to_ipld {
...
+derive_try_from_ipld!(Integer, i64);
+derive_try_from_ipld!(Integer, i128);
+derive_try_from_ipld!(Integer, isize);
+derive_try_from_ipld!(Integer, u8);
+derive_try_from_ipld!(Integer, u16);
+derive_try_from_ipld!(Integer, u32);
+derive_try_from_ipld!(Integer, u64);
+derive_try_from_ipld!(Integer, u128);
+derive_try_from_ipld!(Integer, usize);
+//derive_from_ipld!(Float, f32);
```
vmx: What about adding a comment saying something like "f32 conversion is intentionally not done as it would be a lossy tranformation"

vmx:
> Should I gate the `try_from` with std? Seams required for the error conversion.

Yes please do. I haven't checked if it would theoretical be possible doing it in a no-std environment. But that can then be a separate PR in case it's possible/someone finds the time to tackle it.

vmx:
I've just merged https://github.com/ipld/libipld/pull/180, so Clippy should be good now in case you rebase.

core/src/convert.rs
```diff
@@ -114,46 +113,88 @@ derive_into_ipld!(Map, BTreeMap<String, Ipld>, to_owned);
 derive_into_ipld!(Link, Cid, clone);
 derive_into_ipld!(Link, &Cid, to_owned);
 
+#[cfg(feature = "std")]
```
vmx, line 116: You might be able to do something like (I haven't tried):

```rust
#[cfg(feature = "std")]
{
    derive_try_from_ipld!(Bool, bool);
    derive_try_from_ipld!(Integer, i8);
    …
}
```
which might make things easier to read.

SionoiS: I tried you can't :(

vmx: That sucks.

vmx (approved):
Thanks a lot! This looks really good now.

PR #183, opened 2023-05-24
Remove unused multihash dev-dependency

rillian:
This crate is already specified as part of the library build, with no extra features necessary for test code. Remove the extra reference to minimize the dependencies locked by downstream users.


vmx (approved):
Thanks!

PR #191, opened 2023-09-14
Fix warnings on rustc 1.72

aDotInTheVoid:
Due to `#![deny(warnings)]` these are needed to develop on the project with the latest release.

```
error: using `.deref()` on a double reference, which returns `&T` instead of dereferencing the inner type
  --> core/src/codec.rs:50:13
   |
50 |         self.deref().encode(c, w)
   |             ^^^^^^^^
   |
note: the lint level is defined here
  --> core/src/lib.rs:3:9
   |
3  | #![deny(warnings)]
   |         ^^^^^^^^
   = note: `#[deny(suspicious_double_ref_op)]` implied by `#[deny(warnings)]`
```

```
Error:   --> dag-cbor/src/cbor.rs:74:17
   |
74 |                 20 | 21 | 22 => (),
   |                 ^^^^^^^^^^^^ help: try: `20..=22`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#manual_range_patterns
note: the lint level is defined here
  --> dag-cbor/src/lib.rs:3:9
   |
3  | #![deny(warnings)]
   |         ^^^^^^^^
   = note: `#[deny(clippy::manual_range_patterns)]` implied by `#[deny(warnings)]`
```

```
error: incorrect implementation of `clone` on a `Copy` type
  --> core/src/link.rs:44:29
   |
44 |       fn clone(&self) -> Self {
   |  _____________________________^
45 | |         Self {
46 | |             cid: self.cid,
47 | |             _marker: self._marker,
48 | |         }
49 | |     }
   | |_____^ help: change this to: `{ *self }`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#incorrect_clone_impl_on_copy_type
   = note: `#[deny(clippy::incorrect_clone_impl_on_copy_type)]` on by default
```

```
error: use of `default` to create a unit struct
  --> dag-cbor-derive/examples/basic.rs:55:47
   |
55 |     assert_roundtrip(DagCborCodec, &UnitStruct::default(), &ipld!(null));
   |                                               ^^^^^^^^^^^ help: remove this call to `default`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#default_constructed_unit_structs
   = note: `-D clippy::default-constructed-unit-structs` implied by `-D warnings`
```

```
error: used `unwrap()` on `Some` value
   --> core/tests/serde_deserializer.rs:534:35
    |
[...]

PR #192, opened 2023-09-14
dag-cbor: Implement Encode/Decode for NonZero*

aDotInTheVoid:
Done in the obvious way.

CI failure is spurious due to #191, once that it merged I'll rebase and it'll go green 🤞

### DIFF
--- a/dag-cbor/src/decode.rs
+++ b/dag-cbor/src/decode.rs
@@ -229,6 +229,33 @@ macro_rules! impl_num {
 impl_num!(unsigned u8, u16, u32, u64, u128);
 impl_num!(signed i8, i16, i32, i64, i128);
 
+macro_rules! impl_nonzero {
+    ($(($nzero:ty => $base:ty))*) => {
+        $(
+            impl Decode<DagCbor> for $nzero {
+                fn decode<R: Read + Seek>(c: DagCbor, r: &mut R) -> Result<Self> {
+                    let base = <$base>::decode(c, r)?;
+                    let nzero = <$nzero>::try_from(base)?;
+                    Ok(nzero)
+                }
+            }
+        )*
+    };
+}
+
+impl_nonzero!(
+    (std::num::NonZeroU8 => u8)
+    (std::num::NonZeroU16 => u16)
+    (std::num::NonZeroU32 => u32)
+    (std::num::NonZeroU64 => u64)
+    (std::num::NonZeroU128 => u128)
+    (std::num::NonZeroI8 => i8)
+    (std::num::NonZeroI16 => i16)
+    (std::num::NonZeroI32 => i32)
+    (std::num::NonZeroI64 => i64)
+    (std::num::NonZeroI128 => i128)
+);
+
 impl Decode<DagCbor> for f32 {
     fn decode<R: Read + Seek>(_: DagCbor, r: &mut R) -> Result<Self> {
         // TODO: We don't accept f16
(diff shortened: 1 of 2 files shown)


vmx (commented):
Thanks a lot, the change looks good (waiting for the rebase before approving it).

I'd just like to note that this DAG-CBOR implementation is kind of the legacy one. I recommend everyone looking into https://github.com/ipld/serde_ipld_dagcbor instead.

aDotInTheVoid:
> (waiting for the rebase before approving it)

Rebased

>  I recommend everyone looking into https://github.com/ipld/serde_ipld_dagcbor instead.

I wan't to avoid using serde for this. My use case is implementing the [atproto data model](https://atproto.com/specs/data-model#bytes), which requires different handling of bytes depending on if you're encoding to CBOR or JSON.

Eg:

```json
{
  "exampleBytes": {
    "$bytes": "nFERjvLLiw9qm45JrqH9QTzyC2Lu1Xb4ne6+sBrCzI0"
  }
}
```

Should be encoded in CBOR as an object with key `exampleBytes` and a Byte String (Major 2). I don't think there's a way for serde to know what format it's serializing to, so for now it's easier for everything to derive both `Serialize` and `DagCbor`.

If you've got ideas to work around this, I'd love to hear them.

vmx (approved):

vmx:
Thanks for you contributions. In regards to your question. I don't know what exactly what you're doing, but I tried to come up with something which might help:

```rust
use serde::{ser::SerializeStruct, Deserialize, Deserializer, Serialize, Serializer};

#[derive(Debug, Deserialize, Serialize)]
struct Example {
    #[serde(rename = "exampleBytes")]
    example_bytes: Vec<u8>,
}

#[derive(Debug, Deserialize, Serialize)]
struct ExampleJson {
    #[serde(with = "Wrapper", rename = "exampleBytes")]
    example_bytes: Vec<u8>,
}

impl From<ExampleJson> for Example {
    fn from(other: ExampleJson) -> Self {
        Self {
            example_bytes: other.example_bytes,
        }
    }
}

// Based on https://stackoverflow.com/questions/54657873/is-there-a-way-to-omit-wrapper-root-objects-when-deserializing-objects-with-serd/54658155#54658155
#[derive(Deserialize, Serialize)]
struct Wrapper<T> {
    #[serde(rename = "$bytes")]
    bytes: T,
}

impl<T> Wrapper<T> {
    fn deserialize<'de, D>(deserializer: D) -> Result<T, D::Error>
    where
        T: Deserialize<'de>,
        D: Deserializer<'de>,
    {
        let wrapper = <Self as Deserialize>::deserialize(deserializer)?;
        Ok(wrapper.bytes)
    }

    fn serialize<S>(data: &T, serializer: S) -> Result<S::Ok, S::Error>
    where
        S: Serializer,
        T: Serialize,
    {
        let mut state = serializer.serialize_struct("Wrapper", 1)?;
        state.serialize_field("$bytes", &data)?;
        state.end()
    }
}

fn main() {
    let input_json = r#"{
  "exampleBytes": {
    "$bytes": [1,2,3]
  }
}"#;

    let parsed_input: ExampleJson = serde_json::from_str(input_json).unwrap();
    println!("parsed: {:?}", parsed_input);
    let example: Example = parsed_input.into();
    println!("example: {:?}", example);
    let cbor = serde_ipld_dagcbor::to_vec(&example);
    println!("cbor: {:02X?}", cbor);
}
```

I'm happy to discuss it further if that's not your use case, as I'd really like to get people using `serde_ipld_dagcbor` instead of this DAG-CBOR crate.

vmx:
Or, if that suits you better, feel free to open an issue on the https://github.com/ipld/serde_ipld_dagcbor repo with the problem you are having and we go from there.

PR #199, opened 2024-11-27
chore: use workspace.dependencies

hanabi1224:
(Follow up https://github.com/multiformats/rust-cid/pull/165)
To prepare upgrading `cid` and `multihash`, this PR
- moves all dependencies to under `[workspace.dependency]`
- remove unused dependencies
- bump `thiserror`
- replace `fnv` with `ahash`
- move package metadata to under `[workspace.package]`
- fix or suppress clippy warnings

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,33 +1,31 @@
 [package]
 name = "libipld"
-version = "0.16.0"
-authors = ["David Craven <david@craven.ch>"]
-edition = "2021"
-license = "MIT OR Apache-2.0"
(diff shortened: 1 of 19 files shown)


vmx:
Thanks for the PR. But this repo is going to be deprecated: https://github.com/ipld/libipld/pull/198. As RUSTSEC is pretty aggressive in regards to deprecations, my plan was to update all upstream users I know of to https://github.com/ipld/rust-ipld-core first and then do the actual deprecation.

This means that there shouldn't be any further changes to this library.

PR #201, opened 2025-05-15
Update cid version to v0.11.1 and multihash to v0.19.3

afbase:
relates to #197

Issue #197, opened 2024-03-18
Update to multihash dependency to `0.19` and `cid` dependency to `0.11` 


matheus23:
Is there a path forward for `libipld` to switch to the newer multihash library?

For anyone reading this and interested in updating to `multihash` version `0.19`, you can also switch to [`ipld-core`](https://crates.io/crates/ipld-core), but for anyone else looking for a smoother transition, is updating `libipld` viable?

vmx:
I've there's a problem with upgrading to `ipld-core`, please let me know and I'll have a look. The goal for me is to transition everone away from `libipld`.

PR #202, opened 2025-05-15
Update cid version with allow

afbase:


afbase:
closes #201 
closes #197

